PR #2, opened 2019-05-28
Move provider

cpuguy83:
Sadly this may be difficult to preserve history for... I will have to fiddle with git a bit more.
The approach I took here did not work because it expects a fresh repo.

Closes virtual-kubelet/virtual-kubelet#653


victornoel:
@cpuguy83 for the record you may be able to use `filter-branch` to create a new branch containing the only the subdirectory you are interested in from the other repo, add it as a remote to the current repo, rebase the new branch onto master and then merge it into master.

Actually you don't really need the rebase, git will be quite happy with two branches starting from zero and then merged together :)

cpuguy83:
And now with history :)

cpuguy83:
All green and validated

We should work out e2e (similar to what we have upstream), but this is ok for now since it's just copying over existing code.

PR #4, opened 2019-06-18
Use upstream CLI as lib

cpuguy83:
With providers removed from upstream AND provider registration being
explicit rather than automatic via build tag we can now use the upstream
CLI code as a library except for, of course, the `package main`, which
we want to customize anyway.


cmd/virtual-kubelet/main.go
```diff
@@ -54,17 +55,20 @@ func main() {
 
 	var opts root.Opts
 	optsErr := root.SetDefaultOpts(&opts)
+	opts.Provider = "azure"
```
sargun: Given this always has to be set, it might be nice to do something like newProviderOpts(providerName), which also sets default opts.

cpuguy83: We could if there are other options to do as well.

cpuguy83:
I think this is good. Can always change if it turns out to be problematic due to changes or whatever.

PR #5, opened 2019-06-18
[Carry] Bugfix: Azure ACI provider skips ConfigMap Data exctraction on volumes mapping

cpuguy83:
Carries virtual-kubelet/virtual-kubelet#640

### DIFF
--- a/aci.go
+++ b/aci.go
@@ -1430,6 +1430,9 @@ func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


grahamhayes:
Works as expected:

```
root@wk-caas-3435ee17bffc46e6bcc06a85b3530ca6-49bd8039725a0e7e912b0a:/# cat /kubeless/function.py
def main():
  print "test"
```

From the repro case attached to bugs in vk/vk

cpuguy83:
Thanks for testing!

PR #20, opened 2019-11-09
Add helm chart for deploying VK v1.1 in AKS -- from new registry

ibabou:


macolso:
are we supporting helm 2 or 3?

ibabou:
> are we supporting helm 2 or 3?

I ran it with helm 2.

PR #24, opened 2019-11-16
Add support for test config from env

cpuguy83:
This sets some defaults like before but adds the option to configure
from env, pretty much for CI.


cpuguy83:
I expect this to fail just because the network profile seems like it might be tied to the region and I added a var to circleci for just the region.

cpuguy83:
@ibabou Still getting 409's with eastus.

feiskyer:
Close since long time no activity.

PR #31, opened 2020-01-02
SP warning / az aks install-connector deprecation

macolso:
add warning about SP expiration within a year for manual VK installation 
add warning about az aks install-connector being deprecated


macolso:
added service principal expiration warning - @eugen-nw

PR #33, opened 2020-01-10
VK ACI v1.2.1 changes - Update core + Multiple fixes

ibabou:
The PR contains the following changes:

1. Moving to VK core v1.2.1.
2. Implementing the Async Pod updates API, and handling the status syncing on the provider side.
3. Adjusting the number of workers to allow more operations to be handled in case of large burst volumes of creations.
4. Fix for the NotFound bug where pods gets leaked and re-created unnecessarily, if processing is not quick enough to handle too many CreatePod() operations. With multiple offline tests, now creating a Replicaset of 500 non-VNET alpine-based pods can be completed (up to a running state for all) in less than 4 mins.
5. Adding a cleanup background job (runs periodically every 5 minutes) to delete any leaked container groups if doesn't exist on API-server anymore.
6. Handling the termination, by setting the status of containers correctly after deletion. Hanging "Terminating" state shouldn't happen now if CG is deleted successfully.
7. Fix for the pod state in which it transitions to Ready before having all containers up and in a Ready state, which affects ingress.
8. Fix the handling of dockerconfig secrets in case of private ACR repos, with regards to the additional base encoded user:pass field.
9. Updating supported regions list to reflect the availability on ACI side (that supports VNETs). KoreaCentral is introduced.
10. Minor refactoring & logging enhancements in ACI provider's implementation.


macolso:
lgtm

PR #41, opened 2020-03-13
Update clients to enable VN integration in non prod clouds

ibabou:


MikkelHegn:
Is there an issue describing this change?

ibabou:
> Is there an issue describing this change?

there was an existing issue before. linked it.

PR #53, opened 2020-08-11
fix broken ToC links

chipzoller:


MikkelHegn (approved):

MikkelHegn:
Better late than never... Thanks @chipzoller

PR #59, opened 2020-09-05
update k8sVersion var to match k8s.io/kubernetes v1.18.4 import

nkiraly:
update k8sVersion var to match k8s.io/kubernetes v1.18.4 import update from https://github.com/virtual-kubelet/azure-aci/pull/52


cpuguy83 (approved):
Seems fine, LGTM

PR #60, opened 2020-09-05
correct DNS name annotation domain to project's virtual-kubelet.io

nkiraly:
correct DNS name annotation domain to project's virtual-kubelet.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,7 +550,7 @@ helloworld-2559879000-8vmjw  myResourceGroup    Succeeded            microsoft/a
 <!--
 ### Schedule an ACI pod with a DNS Name label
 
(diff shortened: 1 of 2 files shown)


cpuguy83:
I don't think we can just change this without breaking existing uses.

nkiraly:
I opened this pull request as a follow up to when I first implemented an ACI virtual-kubelet where I got the annotation wrong in my manifests due to the domain inconsistency between the `virtual-kubelet.io/gpu-type` and `virtualkubelet.io/dnsnamelabel` annotations.

I have searched the other providers implementations and did not find references to the dnsnamelabel annotation.

What if you accept the change and release it in a minor, noting the breaking change in release notes ?

The alignment this change brings will avoid both project developers' and adopters' time wasted by removing the chances of the same wrinkle I experienced happening to others going forward.

feiskyer:
Close per the comment https://github.com/virtual-kubelet/azure-aci/pull/60#issuecomment-689686268

Issue #69, opened 2020-09-25
Upgrade Helm charts to use virtualkubelet.azurecr.io/virtual-kubelet:1.3


MikkelHegn:
Following #62 - we should upgrade Helm charts as well.

Issue #62, opened 2020-09-21
Update the azure-aci provider version to 1.3 as it's now using VK 1.3


MikkelHegn:
This PR updated the VK library to 1.3: https://github.com/virtual-kubelet/azure-aci/commit/cadeb6e99263d475a5aa37f320857ca1efab9c9f.

We should bump the aci-provider version to 1.3 as well.

mingqishao:
I have release new version 1.3 with align with Virtual Kubelet.

PR #70, opened 2020-09-25
upgraded Helm chart to v 1.3 and changed owner

MikkelHegn:
Following #62 this is the upgrade to the Helm chart.
Also updated the Helm chart owner to @mingqishao

@mingqishao / @feiskyer - what do you think about the fact that the Helm chart has the latest version, which I've overwritten? Is this a thing we should not do?


MikkelHegn:
Fixes #69

MikkelHegn:
The change in the output path for the tgz file was made to avoid permission issues when running locally. I don't see this target being run as par tof any pipeline.

MikkelHegn:
HOLD merging until I've verified, once a new image with the 1.3.0 tag is ready.

MikkelHegn:
Good to go!

Makefile
```diff
@@ -68,4 +68,4 @@ bin/%:
 helm: bin/virtual-kubelet.tgz
 
 bin/virtual-kubelet.tgz:
-	rm -rf /tmp/virtual-kubelet && mkdir /tmp/virtual-kubelet && cp -r helm/* /tmp/virtual-kubelet/ &&  tar -zcvf bin/virtual-kubelet.tgz -C /tmp virtual-kubelet
+	rm -rf /tmp/virtual-kubelet && mkdir /tmp/virtual-kubelet && cp -r helm/* /tmp/irtual-kubelet/ &&  tar -zcvf virtual-kubelet.tgz -C /tmp virtual-kubelet
```
feiskyer, line 71: `/tmp/irtual-kubelet/` should be `/tmp/virtual-kubelet/`

feiskyer (approved):

PR #92, opened 2020-12-03
Bump node-cli to 0.4.0

cpuguy83:
This adds on-by-default TLS client authentication.
If client certs are not specified then the API server will not be setup.

Full diff: https://github.com/virtual-kubelet/node-cli/compare/v0.3.1..v0.4.0
Release notes: https://github.com/virtual-kubelet/node-cli/releases/tag/v0.4.0

Before releasing with this change, we'll need to work out how to deal with certificates.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/gorilla/websocket v1.4.0
 	github.com/pkg/errors v0.8.1
 	github.com/sirupsen/logrus v1.4.2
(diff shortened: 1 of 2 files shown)


feiskyer (approved):

feiskyer:
@cpuguy83 we could add an option in helm chart for TLS client authentication. Let's add it in a following PR as it needs a new released image.

PR #139, opened 2021-03-16
Enable retry policy for http requests

gossion:


client/client.go
```diff
@@ -128,15 +151,74 @@ func (t userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+			}
+		}
+
+		// The error is likely recoverable so retry. this includes
+		// conection closed, connection failure, timeout, request canceled, etc.
+		return true, nil
+	}
+
+	// refer to https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#general-rest-and-retry-guidelines
+	if resp.StatusCode == http.StatusTooManyRequests ||
```
feiskyer, line 197: define a const for those http status code list?

client/client.go
```diff
@@ -31,8 +39,23 @@ type userAgentTransport struct {
...
+	redirectsErrorRe = regexp.MustCompile(`stopped after \d+ redirects\z`)
+
+	// A regular expression to match the error returned by net/http when the
+	// scheme specified in the URL is invalid. This error isn't typed
+	// specifically so we resort to matching on the error string.
+	schemeErrorRe = regexp.MustCompile(`unsupported protocol scheme`)
+
+	retryWaitMin = 1 * time.Second
+	retryWaitMax = 60 * time.Second
+	retryMax     = 40
```
feiskyer, line 58: could you make those retry configurations configurable, so that they could be tuned easily?

client/client.go
gossion: do you mean adding a const that list all retrayable status code? if yes, I guess I can use https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L48 directly.

client/client.go
gossion: good suggestion.

client/client.go
feiskyer: yes. but we may have different retry status codes here (e.g. cloud provider is using a different list).

gossion: added

client/client.go
gossion: added by reading env

feiskyer (approved):
lgtm

feiskyer:
could you fix the CI failures?

feiskyer (approved):

PR #164, opened 2021-10-14
Add cross-compile support to the Dockerfile

cpuguy83:
This cross compiles the go binary and subsequent image produced when
building with `--platform=<platform>`.

Also adds some build-time optimizations by persisting a cache on the
build machine for the go package cache and the go build cache.


cpuguy83:
I added a new commit that adds a build target specifically to produce an image with the inputs that we need.

PR #169, opened 2021-11-01
Support cross tenant/subscription VNET

mingqishao:
There are two changes was introduced in ACI's new API version 2021-07-01:

1. support create Container Group in difference subscription in Virtual Network 
2. deprecate using NetworkProfile resource

This PR is to corresponding changes in Virtual Kubelet. 
1. Allow customer to specify the Virtual Network's subscription ID by environment variable ACI_VNET_SUBSCRIPTION_ID
2. use the ACI's API 2021-07-01
3. deprecate the NetworkProfile resource. Don't create NetworkProfile  resource before create container.


cpuguy83 (commented):
I don't really have a good way to test this, but code LGTM

client/aci/client_test.go
```diff
@@ -303,21 +303,23 @@ func TestCreateContainerGroupWithBadVNetFails(t *testing.T) {
 					},
 				},
 			},
-			NetworkProfile: &NetworkProfileDefinition{
```
feiskyer, line 306: seems all NetworkProfile related codes have been deleted, would this cause issues?

mingqishao: As I mentioned in RP's description, the NetworkProfile has been deprecated in ACI's new API version 2021-07-01

PR #174, opened 2021-11-15
Support real-time metrics

mingqishao:
### Why this PR
1. adding real-time metrics support.  VK use the ACI's Real-Time Metrics Extension as data source for stats/summary  API. That will improve the metrics latency in Metrics Server from 1~2 minutes to 5 ~ 10 seconds.
2. Fix some issues in Container Insights data source based metrics API. The main problem is the value of CPU.UsageCoreNanoSeconds and Network should be cumulative value, instead of average values.
3. refactor the metrics part codes of ACI Provider.

### adding real-time metrics support.
1. Add a new system environment 'ENABLE_REAL_TIME_METRICS' that allow VK turn on the real-time metrics features. Real-time Metrics Extension will be added in Container Group's spec if this feature be enabled.
2. Implementation of this features are in "provider/metrics/real_time.go". The logic is straightforward: retrieving metrics data by calling real-time metrics API on http://<pod-ip>:18899/v1/stats

### Fix some issues in original Container Insights metrics API
1. All Container Insights metrics be implemented in "provider/metrics/container_insights.go"
2. calculate the cumulative value for CPU.UsageCoreNanoSeconds and Network from Container Insights's average value.

### refactor the metrics part of ACI Provider.
1. create a separate package "provider/metrics" for PodMetricsProvider. 
2. implement the metric cache logic in "provider/metrics/cache.go"
3. implement a "podStatsGetterDecider" in "provider/metrics/metrics.go" which to decide using Container Insight or Real-Time Metrics Extension as the metrics data source. Real-time will be used if there are Real-Time Metrics Extension definition in Container Group's spec.


cpuguy83 (changes requested):
I haven't checked your calculations yet.
In general this looks ok, some minor feedback.

Can you add an e2e test case? This way we can at least validate we aren't getting stale stats.

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,268 @@
...
+				TxBytes:  newUInt64Pointer(0),
+				TxErrors: newUInt64Pointer(0),
+			},
+		},
+		Containers: make([]stats.ContainerStats, 0),
+	}
+
+	for _, c := range pod.Status.ContainerStatuses {
+		containerName := c.Name
+		logger.Debugf("star to populate stats for container '%s'", containerName)
```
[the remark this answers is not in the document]
mingqishao: You are right. This is not necessary. I deleted that.

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,268 @@
...
+			stat.Memory.WorkingSetBytes = &podMem
+			stat.Memory.Time = metav1.NewTime(data.Timestamp)
+		}
+		stat.Containers = append(stat.Containers, cs)
+	}
+
+	return stat
+}
+
+func findTimeSeries(metricsResult *aci.ContainerGroupMetricsResult, metricsType aci.MetricType, containerName *string) ([]aci.TimeSeriesEntry, bool) {
```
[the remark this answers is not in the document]
mingqishao: fixed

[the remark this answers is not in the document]
mingqishao: Yes, you are right. The codes of invoking Container Insight is not clear. I have encapsulated Container Insights related codes into containerInsightsMetricsWrapper, and as you suggested, I cached the result to avoid iterate multiple times.

provider/metrics/metrics.go
```diff
@@ -0,0 +1,222 @@
...
+	var errGroup errgroup.Group
+	chResult := make(chan stats.PodStats, len(pods))
+
+	sema := make(chan struct{}, 10)
+	for _, pod := range pods {
+		if pod.Status.Phase != v1.PodRunning {
+			continue
+		}
+		pod := pod
+		errGroup.Go(func() error {
```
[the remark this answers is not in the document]
mingqishao: You are right. That not make sense one Pod's failure will lead to cancel entire thing. I will do fix that in another RP later.

mingqishao:
>

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+type cumulativeUsage struct {
+	lastUpdateTime time.Time
+	value          uint64
+}
+
+func NewContainerInsightsMetricsProvider(metricsGetter ContainerGroupMetricsGetter, resourceGroup string) *containerInsightsPodStatsGetter {
+	return &containerInsightsPodStatsGetter{
+		metricsGetter:        metricsGetter,
+		resourceGroup:        resourceGroup,
+		cumulativeUsageCache: cache.New(time.Minute*30, time.Minute*30),
```
cpuguy83, line 43: Ditto on LRU cache.

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+}
+
+func (containerInsights *containerInsightsPodStatsGetter) getPodStats(ctx context.Context, pod *v1.Pod) (*stats.PodStats, error) {
```
cpuguy83, line 47: Please use single letter names for these objects.
e.g.

```
func (g *containerInsightsPodStatsGetter) getPodStats(ctx context.Context, pod *v1.Pod) (*stats.PodStats, error) {
```

We can probably also shorten `containerInsightsPodStatsGetter` to `insightsGetter` or maybe `getterV0`

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+
+func (containerInsights *containerInsightsPodStatsGetter) updateCumulativeValues(logger log.Logger, pod *v1.Pod, metrics *containerInsightsMetricsWrapper, stat *stats.PodStats) {
+	cacheKey := string(pod.UID)
+	c, found := containerInsights.cumulativeUsageCache.Get(cacheKey)
+	var cachedCumulativeValue *podCumulativeUsage
+	if !found {
+		logger.Infof("Didn't find cumulative values in cache for pod: pod name '%s', UID '%s'", pod.Name, cacheKey)
+		cachedCumulativeValue = &podCumulativeUsage{
+			podUID: cacheKey,
+			networkRx: cumulativeUsage{
```
cpuguy83, line 154: These initializations are redundant.

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+				lastUpdateTime: time.Time{},
+				value:          0,
+			},
+			networkTx: cumulativeUsage{
+				lastUpdateTime: time.Time{},
+				value:          0,
+			},
+			containersUsageCoreNanoSeconds: make(map[string]cumulativeUsage),
```
cpuguy83, line 162: Except this one, this is not redundant

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+			wrapper.networkRxMetricSeries = metrics.Timeseries[0].Data
+		}
+		if metrics.Desc.Value == aci.MetricTyperNetworkBytesTransmittedPerSecond && len(metrics.Timeseries) > 0 {
+			wrapper.networkTxMetricSeries = metrics.Timeseries[0].Data
+		}
+	}
+	return wrapper, nil
+}
+
+func (wrapper *containerInsightsMetricsWrapper) getCPUMetricsSeries(containerName string) ([]aci.TimeSeriesEntry, bool) {
```
cpuguy83, line 299: `getGPU`

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+	val, found := wrapper.cpuMetricsSeries[containerName]
+	return val, found
+}
+
+func (wrapper *containerInsightsMetricsWrapper) getMemoryMetricsSeries(containerName string) ([]aci.TimeSeriesEntry, bool) {
```
cpuguy83, line 304: `getMemory`

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+	val, found := wrapper.memoryMetricsSeries[containerName]
+	return val, found
+}
+
+func (wrapper *containerInsightsMetricsWrapper) getNetworkTxMetricsSeries() []aci.TimeSeriesEntry {
```
cpuguy83, line 309: `getNetworkTx`

provider/metrics/container_insights.go
```diff
@@ -0,0 +1,315 @@
...
+	return wrapper.networkTxMetricSeries
+}
+
+func (wrapper *containerInsightsMetricsWrapper) getNetworkRxMetricsSeries() []aci.TimeSeriesEntry {
```
cpuguy83, line 313: `getNetworkRx`

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	GetPods() []*v1.Pod
+}
+
+// package dependency: query the Pod's correspoinding Container Group metrics from Container Insights
+type ContainerGroupMetricsGetter interface {
+	GetContainerGroupMetrics(ctx context.Context, resourceGroup, containerGroup string, options aci.MetricsRequest) (*aci.ContainerGroupMetricsResult, error)
+}
+
+// package dependency: query the Container Group information
+type ContaienrGroupGetter interface {
```
cpuguy83, line 35: Typo

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	ContainerGroupCacheTTLSeconds = 60 * 5
+)
+
+// package dependency: query the Pods in current virtual nodes. it usually is ResourceManager
+type PodGetter interface {
+	GetPods() []*v1.Pod
+}
+
+// package dependency: query the Pod's correspoinding Container Group metrics from Container Insights
+type ContainerGroupMetricsGetter interface {
```
cpuguy83, line 30: `MetricsGetter`

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	GetContainerGroupMetrics(ctx context.Context, resourceGroup, containerGroup string, options aci.MetricsRequest) (*aci.ContainerGroupMetricsResult, error)
```
cpuguy83, line 31: `GetMetrics`

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+
+func NewACIPodMetricsProvider(nodeName, aciResourcegroup string, podGetter PodGetter, aciCGGetter ContaienrGroupGetter, aciCGMetricsGetter ContainerGroupMetricsGetter) *ACIPodMetricsProvider {
+	provider := ACIPodMetricsProvider{
+		nodeName:           nodeName,
+		podGetter:          podGetter,
+		aciCGGetter:        aciCGGetter,
+		aciCGMetricsGetter: aciCGMetricsGetter,
+	}
+
+	// containerInsightGetter := WrapCachedPodStatsGetter(
```
cpuguy83, line 64: Deleted commented code

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	containerInsightGetter := NewContainerInsightsMetricsProvider(aciCGMetricsGetter, aciResourcegroup)
+	realTimeGetter := WrapCachedPodStatsGetter(
+		5,
+		NewRealTimeMetrics())
+	provider.podStatsGetter = NewPodStatsGetterDecider(containerInsightGetter, realTimeGetter, aciResourcegroup, aciCGGetter)
+	return &provider
+}
+
+// GetStatsSummary returns the stats summary for pods running on ACI
+func (provider *ACIPodMetricsProvider) GetStatsSummary(ctx context.Context) (summary *stats.Summary, err error) {
```
cpuguy83, line 76: `func (p *ACIPodMetricsProvider)`

e2e/pods_test.go
```diff
@@ -7,7 +7,10 @@ import (
 )
 
 func TestPodLifecycle(t *testing.T) {
-	spec, err := fixtures.ReadFile("fixtures/nginx.yml")
+	// delete the pod first
+	kubectl("delete", "pod/vk-e2e-hpa")
+
+	spec, err := fixtures.ReadFile("fixtures/hpa.yml")
```
cpuguy83, line 13: If we aren't going to use the other fixture, can you delete it?

e2e/pods_test.go
```diff
@@ -23,12 +26,30 @@ func TestPodLifecycle(t *testing.T) {
...
 	if out, err := cmd.CombinedOutput(); err != nil {
 		t.Fatal(string(out))
 	}
+	t.Log("success create pod")
+
+	// query metrics
+	deadline = time.Now().Add(5 * time.Minute)
+	for {
+		t.Log("query metrics ....")
+		cmd = kubectl("get", "--raw", "/apis/metrics.k8s.io/v1beta1/namespaces/vk/pods/vk-e2e-hpa")
```
cpuguy83, line 39: Instead of this use of deadline you could change `kubectl` to take a context that has a timeout set and we can bound that to the exec call (`exec.CommandContext` instead of `exec.Command`)

cpuguy83 (changes requested):
Let's try to keep this as go-idomatic as possible, so keep var/object names short, the less important something is the shorter a name should be. Usually the var for a function bounding is a single character, e.g. `func (s *specialThing)` instead of `func (special *specialThing)`

I'm not sure I understand the "decider" thing vs initializing the correct type to begin with.
Maybe I missed a thing here?

Caching. Caching is hard to get right and we should do as little of it as possible.
We should make sure not to have an unbounded cache. Note that maps in go can never shrink in size w/o recreating them and these caches all use maps. If we load tons of pods onto one instance we are going to allocate a lot of memory just for stats.
Generally I think we should only bother caching the result (k8s `StatsSummary`) rather than intermediate values.
We may need to change that at some point since I know metrics-server is moving to a different API... maybe when we add support we can introduce our own intermediate type or maybe `StatsSummary` is still good enough and can be converted appropriately.
But in any case today we only need `StatsSummary`.
I think the big thing here is making sure our stats calculations aren't doing more than they need to (like `for { ... for { ... } }` gets expensive) and utilize caching locally within the scope of a function call rather than globally. If we need to do more we can look at that.
What we don't want to do is make an API call to Azure when it would not serve us any new information anyway and this is where a good global-ish cache comes in handy, but its also not the end of the world if we do have to make an API call.

provider/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+
+type podStatsGetterDecider struct {
+	containerInsightsGetter podStatsGetter
+	realTimeGetter          podStatsGetter
+	rgName                  string
+	aciCGGetter             ContaienrGroupGetter
+	cache                   *cache.Cache
+}
+
+func NewPodStatsGetterDecider(containerInsightsGetter podStatsGetter, realTimeGetter podStatsGetter, rgName string, aciCGGetter ContaienrGroupGetter) *podStatsGetterDecider {
```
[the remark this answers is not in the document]
mingqishao: We can't do that.  The ACI deploy a sidecar in customer's Pod to provide the metrics data. Not all Pod has this sidecar. The pod don't have the sidecar if the Pod was created before VK upgrade to new version which support real-time metrics.  So the Deicer where to query the Container Group's spec to check if CG has the sidecar, then decide use the real-time metrics or Container Insight metrics.

provider/metrics/mock_metrics_test.go
```diff
@@ -0,0 +1,167 @@
...
+// Source: .\metrics.go
+
+// Package mock_metrics is a generated GoMock package.
+package metrics
+
+import (
+	context "context"
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
```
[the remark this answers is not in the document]
mingqishao: I don't like gomock either, especially that require to auto generated codes.  I initially to handle the mock myself, till I aware I need to record how much time be called for test the cache case. So gomock will be much easier than do it myself.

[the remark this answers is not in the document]
mingqishao: by the way, that not a new package. We already use gomaock in other package.

provider/metrics/cache.go
```diff
@@ -0,0 +1,37 @@
...
+	return &cachePodStatsGetter{
+		wrappedGetter: getter,
+		cache:         cache.New(time.Duration(ttlSeconds)*time.Second, 10*time.Minute),
+	}
+}
+
+//Adding cache capability into podStatsGetter
+type cachePodStatsGetter struct {
+	wrappedGetter podStatsGetter
+	cache         *cache.Cache
```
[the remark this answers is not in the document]
mingqishao: There are two scenarios to use cache in this metrics:
1. first is here. cache the result of getPodStats to avoid frequent ACI's API calls. 
2. second is to cache the cumulative value. The new cumulative value should be the old one + delta value.

Both of scenarios just need to:
1. cache the latest value.
2. automatically clean the old data if cache not be used for a while because corresponding Pod has been deleted, or refresh the cumulative data if it not update for a while. 

The [patrickmn/go-cache](https://github.com/patrickmn/go-cache) is what I need. Its codes is straightforward and have the only features I need.   I didn't know the golang-lru library. It look work for first scenario as well. But there is small problem the second scenario. We don't need LRU algorithm. I hope the cache data always there, otherwise the cumulative value will be reset.

provider/metrics/container_insights.go
mingqishao: Here is the second scenarios.  We cache the value for calculating cumulative value. we don't expect the cache be clean if cache run out of capacity.

provider/metrics/container_insights.go
mingqishao: I did that specifically although I know that is redundant. Because:
1. Make sure every filed be explicitly initialized, instead of rely on Go's default value. 
2. That will be easier for reading codes, especially for the reader without too much knowledge about the Golang. Actually I had same issue when I had not too much Golang knowledge.

provider/metrics/container_insights.go
cpuguy83: Because we don't need the name repeated everywhere.
If we put a doc string on the object like `// Implements PodStatsGetter for Container Insights` it puts that important information where the object is defined and reduces the noise everywhere else.
It's also just idiomatic go.

provider/metrics/container_insights.go
cpuguy83: Maybe some docs on the cumulative value.
Are you saying the value can't be recalculated?

provider/metrics/container_insights.go
mingqishao: There are many information in Container Insight's result: metrics's meta data, time series's meta data and multiple time series data.  Here is to get the time series data for CPU. So the long name "getCPUMetricsSeries" more meaningful and easier for understanding the codes.

provider/metrics/container_insights.go
mingqishao: Same as above

provider/metrics/container_insights.go
mingqishao: Same as above

provider/metrics/container_insights.go
mingqishao: Same as above

provider/metrics/metrics.go
mingqishao: As I mentioned before. The longer name should be more meaningful and easier for reading by some else.

provider/metrics/metrics.go
mingqishao: This name is to match with the implementation one in client\aci\metrics.go.  The purpose of the Interface here just for mocking in unit test.

provider/metrics/metrics.go
cpuguy83: Am I not someone else?
For me, the code is wordy enough that I'm having a difficult time to get through it (I have reading issues in general).
Also, these objects/interfaces are already within the context of this metrics package within the ACI provider.

That said I'm only making suggestions but am definitely not nit-picking.

provider/metrics/metrics.go
cpuguy83: I see.
Perhaps we are abstracting the wrong thing here?
We are exporting an interface so that we can provide a mocked implementation for unit tests for something that otherwise only has a single implementation.

Don't want to hold up the PR on such semantics.

provider/metrics/metrics.go
mingqishao: good catch. I comment the lines for diagnose a bug. I forget push them back. I have fixed that.

e2e/pods_test.go
mingqishao: deleted unused fixture.

e2e/pods_test.go
mingqishao: yes, good to know the exec.CommandContext. But for this case, the command return failure or success immediately. I need to retry the command till it return the success. I think what I need is like [retry-go](https://github.com/avast/retry-go) library. But it not worth introduce a new package.

cpuguy83: Oh right, silly me this is for the loop not the individual command.

provider/metrics/container_insights.go
mingqishao: for example the cumulative value for CPU usage. the value should be totally CPU usage from container start up.  If we want to recalculated it, we need query the metrics data start from time point when container created every time. To avoid do that, we cached the value and just add the new value on top of that.

provider/metrics/container_insights.go
mingqishao: Agree with you. I fixed the long name.

provider/metrics/container_insights.go
mingqishao: fixed that.

provider/metrics/container_insights.go
mingqishao: fixed that.

provider/metrics/container_insights.go
mingqishao: fixed that.

provider/metrics/container_insights.go
mingqishao: fixed that.

provider/metrics/metrics.go
mingqishao: fixed.

provider/metrics/metrics.go
mingqishao: agree with you. I fixed that.

provider/metrics/metrics.go
mingqishao: fixed that.

Issue #186, opened 2022-02-28
Improve unit test coverage of client/api/api.go


minhng22:
PR [https://github.com/virtual-kubelet/azure-aci/pull/188](url)

PR #188, opened 2022-03-02
Unit test for client/api/api.go at 100% coverage

minhng22:
Add unit tests to 100% coverage and minor var renaming.

https://github.com/virtual-kubelet/azure-aci/issues/186

### DIFF
--- a/client/api/api.go
+++ b/client/api/api.go
@@ -2,9 +2,7 @@
 package api
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 )
 
@@ -43,21 +41,23 @@ type errorReply struct {
 // CheckResponse returns an error (of type *Error) if the response
 // status code is not 2xx.
 func CheckResponse(res *http.Response) error {
-	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+	const MinValidStatusCode = 200
+	const MaxValidStatusCode = 299
+
+	if res.StatusCode >= MinValidStatusCode && res.StatusCode <= MaxValidStatusCode {
 		return nil
 	}
-
-	slurp, err := ioutil.ReadAll(res.Body)
-	if err == nil {
-		jerr := new(errorReply)
-		err = json.Unmarshal(slurp, jerr)
-		if err == nil && jerr.Error != nil {
-			if jerr.Error.StatusCode == 0 {
-				jerr.Error.StatusCode = res.StatusCode
+	resBodyReads, ioReaderErr := ioReadAll(res.Body)
+	if ioReaderErr == nil {
+		jsonResBody := new(errorReply)
+		jsonUnMarshallErr := jsonUnmarshall(resBodyReads, jsonResBody)
+		if jsonUnMarshallErr == nil && jsonResBody.Error != nil {
+			if jsonResBody.Error.StatusCode == 0 {
+				jsonResBody.Error.StatusCode = res.StatusCode
 			}
-			jerr.Error.Body = string(slurp)
-			jerr.Error.URL = res.Request.URL.String()
-			return jerr.Error
+			jsonResBody.Error.Body = string(resBodyReads)
+			jsonResBody.Error.URL = res.Request.URL.String()
+			return jsonResBody.Error
 		}
 	}
 
--- a/client/api/api_test.go
+++ b/client/api/api_test.go
@@ -1,0 +1,224 @@
+package api
+
+import (
+	. "github.com/onsi/gomega"
+	"io"
+	"math/rand"
+	"net/http"
+	"net/url"
+	"testing"
+)
+
+func TestErrorWithNoMessage(t *testing.T) {
+	// assemble
+	goOmegaInstance := NewWithT(t)
+
+	var errWithNoMessage = Error{
+		StatusCode: rand.Int(),
+		Code:       RandStr(),
+		Body:       RandStr(),
+		Header:     http.Header{"": make([]string, 1)},
+		URL:        RandStr(),
+	}
+
+	// act
+	err := errWithNoMessage.Error()
+
+	// assert
+	goOmegaInstance.Expect(err).To(BeAssignableToTypeOf(""))
+}
+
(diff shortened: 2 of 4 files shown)


client/api/api.go
```diff
@@ -2,12 +2,13 @@
 package api
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 )
 
+const MinValidStatusCodeRange = 200
```
ryanzhang-oss, line 9: do they need to be public?

client/api/api.go
```diff
@@ -42,22 +43,25 @@ type errorReply struct {
 
 // CheckResponse returns an error (of type *Error) if the response
 // status code is not 2xx.
-func CheckResponse(res *http.Response) error {
-	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+func CheckResponse(
+	res *http.Response,
+	ioReader IoReaderUtils,
```
ryanzhang-oss, line 48: In general, tests should not change the signature of the function it tests.  You can simply replace the  line
```
ioutil.ReadAll(res.Body)
```
to be a function all
```
var readAll func(r io.Reader) ([]byte, error)
readAll(res.Body)
```

ryanzhang-oss: Same for the other parameter

ryanzhang-oss: You can also declare them as interfaces and mock the interface

client/api/api.go
tangoat2019: These const are used in the test file.

client/api/api.go
tangoat2019: These dependencies are call from within this function. I think dependency injection style makes it easier to test and maintain. I agreed when we write test we shouldn't change the original function, but I don't think we can mock something that's called within the tested function do we?

client/api/api_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"io"
+	"math/rand"
+	"net/http"
+	"net/url"
+	"testing"
+)
+
+func TestErrorWithNoMessage(t *testing.T) {
+	// assemble
+	goOmegaInstance := NewWithT(t)
```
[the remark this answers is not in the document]
ryanzhang-oss: I see. That's a mix is Ginko lib with vanilla go test framework. I've never seen that before but it's ok.

client/api/api.go
ryanzhang-oss: can you point me to the doc on what does "Go testing guideline" say about the Dependency injection style?  I don't normally see this type of "injection" in Go.

ryanzhang-oss: The reason I ask is that Go is not an OO language so I have never seen any "dependency injection" in this PR.

client/api/api_test.go
```diff
@@ -0,0 +1,230 @@
...
+		Body:       RandStr(),
+		Header:     http.Header{"": make([]string, 1)},
+		URL:        RandStr(),
+	}
+
+	// act
+	err := errWithNoMessage.Error()
+
+	// assert
+	goOmegaInstance.Expect(err).To(BeAssignableToTypeOf(""))
```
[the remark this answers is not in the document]
ryanzhang-oss: similar or exact the same?

client/api/api.go
minhng22: I think I saw that in the Go testing course Sean shared. The goal of that style was to isolate the dependencies. Your way would actually achieve the same purpose. I have never worked with Go but with web API developments this style is very popular and language isn't really the case (I saw this practice common in both NodeJS and Spring). Both those frameworks have built-in support for that though, so using a struct in Go was just my hack. But with that being said, both seems to achieve the same purpose so I am fine with switching style.

client/api/api_test.go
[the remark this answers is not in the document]
minhng22: Identical. The only difference in both case is that the testing function would return a string with different content. Thus the assert is identical, because testing for the content of the message is unstable.

client/api/api.go
ryanzhang-oss: The solution I showed in the first comment does not require the test to change the interface of the function it tests. The trick is to replace the function call from a direct call to a library to a call to a local function.  Please try it out and you will see how it works

minhng22: Synced with Ryan offline and this should be reflected in the PR now

client/api/api.go
```diff
@@ -43,21 +41,23 @@ type errorReply struct {
 // CheckResponse returns an error (of type *Error) if the response
 // status code is not 2xx.
 func CheckResponse(res *http.Response) error {
-	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+	const MinValidStatusCode = 200
```
ryanzhang-oss, line 44: why define a const here?

ryanzhang-oss: why make them pubic?

minhng22: Those are inside CheckResponse so they are not public. The same values are used when testing against this function so it's easier to maintain and read to define them as const.

client/api/api_test.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func TestCheckResponseWithStatusCode2XX(t *testing.T) {
+	// assemble
+	goOmegaInstance := NewWithT(t)
+	const MinValidStatusCode = 200
```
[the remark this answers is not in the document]
minhng22: These const aren't reused anywhere beside the testing functions so I don't make them public and don't reuse them.

client/api/api.go
ryanzhang-oss: You are right, they are not public. However, they are not used anywhere else, are they?

ryanzhang-oss: in this case, why make the change here then?

minhng22: It's easier to read to have a defined const instead of a raw value to me. It's personal preference I don't have strong opinion about it.

minhng22: Synced with Ryan offline and this doesn't seems to be an issue so I would resolve this conversation.

PR #192, opened 2022-04-05
Add ACI Spot virtual nodes support to virtual kubelet

telotlik:
This change adds ACI spot virtual node support to virtual kubelet. 

The user can add ACI priority ('Spot' or 'Regular') to the pod definition as a part of Annotations, which will be used to direct the pod to run either on ACI Spot pool virtual nodes (for Spot priority) or ACI pool virtual nodes (for Regular priority).
This change enables users to add Spot Virtual Node to existing AKS cluster or create a new cluster which has Spot Virtual Nodes. All pods deployed on Spot Virtual Nodes will be run on ACI Spot Containers.

Validated the change by testing locally on minikube as well as by deploying on AKS cluster. Also added tests to verify that the pod is created only when valid values are passed for priority property in annotations.

### DIFF
--- a/provider/aci.go
+++ b/provider/aci.go
@@ -657,6 +657,16 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	containerGroup.Location = p.region
 	containerGroup.RestartPolicy = aci.ContainerGroupRestartPolicy(pod.Spec.RestartPolicy)
 	containerGroup.ContainerGroupProperties.OsType = aci.OperatingSystemTypes(p.operatingSystem)
+	priority := pod.Annotations["priority"]
+	if priority != "" {
+		if priority == "Spot" {
+			containerGroup.ContainerGroupProperties.Priority = aci.Spot
+		} else if priority == "Regular" {
+			containerGroup.ContainerGroupProperties.Priority = aci.Regular
+		} else {
+			return fmt.Errorf("The pod requires either Regular or Spot priority. Invalid value %s", priority)
+		}
+	}
 
 	// get containers
 	containers, err := p.getContainers(pod)
@@ -716,6 +726,7 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 		"Namespace":         pod.Namespace,
 		"UID":               podUID,
 		"CreationTimestamp": podCreationTimestamp,
+		"Priority":          pod.Annotations["priority"],
 	}
 
 	p.amendVnetResources(&containerGroup, pod)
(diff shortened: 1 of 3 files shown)


provider/aci.go
```diff
@@ -657,6 +661,19 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
...
 	containerGroup.ContainerGroupProperties.OsType = aci.OperatingSystemTypes(p.operatingSystem)
+	if pod.Annotations != nil {
+		priority, priorityExists := pod.Annotations[priorityTypeAnnotation]
+		if priorityExists {
+
+			if strings.EqualFold(priority, string(aci.Spot)) {
+				containerGroup.ContainerGroupProperties.Priority = aci.Spot
+			} else if strings.EqualFold(priority, string(aci.Regular)) {
+				containerGroup.ContainerGroupProperties.Priority = aci.Regular
+			} else {
```
ryanzhang-oss, line 672: won't this fail all the existing ones?

provider/aci.go
```diff
@@ -61,6 +61,10 @@ const (
 	gpuTypeAnnotation = "virtual-kubelet.io/gpu-type"
 )
 
+const (
+	priorityTypeAnnotation = "priority"
```
mlalic, line 65: This string should be prefixed in some way in order to not conflict with potential annotations already in use by a customer pod for maximum compatibility with customers' existing pods. I suspect `priority` is a reasonably frequent "private" annotation. 

See for example `virtualKubeletDNSNameLabel` or `gpuTypeAnnotation`. 

I'd say that an argument can be made that given how ACI-specific this is, we could "namespace" it even more specifically than just virtual kubelet, but at the very least, IMO, it should match the other annotation-based extensions here.

provider/aci.go
```diff
@@ -728,6 +738,26 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	return p.createContainerGroup(ctx, pod.Namespace, pod.Name, &containerGroup)
 }
 
+// Set the Container Group Priority Property
+// Accepted Values : Regular, Spot
```
mlalic, line 742: Since the priority is not an explicit parameter to this function, this comment should explain how priority is passed in, if we already mention what the accepted values are :) 

Additionally, as far as I can tell from the code, it is also allowed not to specify it at all, so I'd recommend mentioning that too.

provider/aci.go
telotlik: Validated that this change does not affect existing behavior

helayoty:
@telotlik please rebase and make sure all pipelines pass.

PR #193, opened 2022-04-15
feat: Add support for CSI driver

### LINKED ISSUE #88: [feature request] support Azure File CSI driver
Current azure-aci only supports in-tree azure file driver:
https://github.com/virtual-kubelet/azure-aci/blob/190d6b3e255d558c2b0a8dc9c152f1455e172439/provider/aci.go#L1611-L1631

also need to support [Azure File CSI driver](), in the near future, in-tree driver will be removed, and only CSI driver will be supported, here is the reference code about how to get secret, file share from CSI driver struct:
https://github.com/kubernetes/csi-translation-lib/blob/833303372de1d3df6c92110443c70a4c61b7aa50/plugins/azure_file.go#L167-L174

helayoty:
This PR will add the support of the AzureFile CSI driver in addition to the unit test.
 
Fixes #88

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.6.0
-	github.com/stretchr/testify v1.5.1
 	github.com/thoas/go-funk v0.9.1
 	github.com/virtual-kubelet/node-cli v0.7.0
 	github.com/virtual-kubelet/virtual-kubelet v1.6.0
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.6.0
+	github.com/stretchr/testify v1.5.1
 	github.com/thoas/go-funk v0.9.1
 	github.com/virtual-kubelet/node-cli v0.7.0
 	github.com/virtual-kubelet/virtual-kubelet v1.6.0
--- a/provider/aci.go
+++ b/provider/aci.go
@@ -1678,16 +1678,16 @@ func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	for _, v := range pod.Spec.Volumes {
 		// Handle the case for Azure File CSI driver
 		if v.CSI != nil {
-			// Check if the CSI driver if file (Disk is not supported by ACI)
+			// Check if the CSI driver is file (Disk is not supported by ACI)
 			if v.CSI.Driver == AzureFileDriverName {
 				csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
 				if err != nil {
-					return volumes, err
+					return nil, err
 				}
 				volumes = append(volumes, *csiVolume)
 				continue
 			} else {
-				return nil, fmt.Errorf("Pod %s requires volume %s which is of an unsupported type", pod.Name, v.Name)
+				return nil, fmt.Errorf("Pod %s requires volume %s which is of an unsupported type %s", pod.Name, v.Name, v.CSI.Driver)
 			}
 		}
 
--- a/provider/aci.go
+++ b/provider/aci.go
@@ -7,7 +7,6 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +17,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/gorilla/websocket"
 	client "github.com/virtual-kubelet/azure-aci/client"
@@ -50,6 +51,8 @@ const (
 	// Parameter names defined in azure file CSI driver, refer to
 	// https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md
 	azureFileShareName = "shareName"
+	// AzureFileDriverName is the name of the CSI driver for Azure File
+	AzureFileDriverName = "file.csi.azure.com"
 )
 
 // DNS configuration settings
@@ -1644,7 +1647,7 @@ func (p *ACIProvider) getAzureFileCSI(volume v1.Volume, namespace string) (*aci.
 		// Set the storage account name and key
 		secret, err := p.resourceManager.GetSecret(volume.CSI.NodePublishSecretRef.Name, namespace)
 		if err != nil {
-			return nil, fmt.Errorf("The secret %s for AzureFile CSI driver %s is not found, err: %s", volume.CSI.NodePublishSecretRef.Name, volume.Name, err)
+			return nil, errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found", volume.CSI.NodePublishSecretRef.Name, volume.Name))
 		}
 
 		if secret == nil {
@@ -1675,12 +1678,17 @@ func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	for _, v := range pod.Spec.Volumes {
 		// Handle the case for Azure File CSI driver
 		if v.CSI != nil {
-			csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
-			if err != nil {
-				return volumes, err
+			// Check if the CSI driver if file (Disk is not supported by ACI)
+			if v.CSI.Driver == AzureFileDriverName {
+				csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
+				if err != nil {
+					return volumes, err
+				}
+				volumes = append(volumes, *csiVolume)
+				continue
+			} else {
+				return nil, fmt.Errorf("Pod %s requires volume %s which is of an unsupported type", pod.Name, v.Name)
 			}
-			volumes = append(volumes, *csiVolume)
-			continue
 		}
 
 		// Handle the case for the AzureFile volume.
--- a/provider/aci.go
+++ b/provider/aci.go
@@ -47,6 +47,9 @@ const (
 	virtualKubeletDNSNameLabel = "virtualkubelet.io/dnsnamelabel"
 
 	subnetDelegationService = "Microsoft.ContainerInstance/containerGroups"
+	// Parameter names defined in azure file CSI driver, refer to
+	// https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md
+	azureFileShareName = "shareName"
 )
 
 // DNS configuration settings
@@ -1632,9 +1635,54 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
 	}, nil
 }
 
+func (p *ACIProvider) getAzureFileCSI(volume v1.Volume, namespace string) (*aci.Volume, error) {
+	azureSource := &aci.AzureFileVolume{
+		ReadOnly: *volume.CSI.ReadOnly,
+	}
+	if volume.CSI.NodePublishSecretRef != nil && volume.CSI.NodePublishSecretRef.Name != "" {
+
+		// Set the storage account name and key
+		secret, err := p.resourceManager.GetSecret(volume.CSI.NodePublishSecretRef.Name, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("The secret %s for AzureFile CSI driver %s is not found, err: %s", volume.CSI.NodePublishSecretRef.Name, volume.Name, err)
+		}
+
+		if secret == nil {
+			return nil, fmt.Errorf("Getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
+		}
+
+		azureSource.StorageAccountName = string(secret.Data["azurestorageaccountname"])
+		azureSource.StorageAccountKey = string(secret.Data["azurestorageaccountkey"])
+
+		// Set shareName
+		if volume.CSI.VolumeAttributes != nil {
+			if shareName, ok := volume.CSI.VolumeAttributes[azureFileShareName]; ok {
+				azureSource.ShareName = shareName
+			}
+		}
+		volume := aci.Volume{
+			Name:      volume.Name,
+			AzureFile: azureSource,
+		}
+		return &volume, nil
+	} else {
+		return nil, fmt.Errorf("NodePublishSecretRef for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
+	}
+}
+
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
+			if err != nil {
+				return volumes, err
+			}
+			volumes = append(volumes, *csiVolume)
+			continue
+		}
+
 		// Handle the case for the AzureFile volume.
 		if v.AzureFile != nil {
 			secret, err := p.resourceManager.GetSecret(v.AzureFile.SecretName, pod.Namespace)
--- a/provider/aci_test.go
+++ b/provider/aci_test.go
@@ -1427,7 +1427,11 @@ func TestCreatePodWithCSIVolume(t *testing.T) {
 
 			err = provider.CreatePod(context.Background(), pod)
 
-			assert.Equal(t, tc.expectedError, err)
+			if tc.expectedError != nil {
+				assert.Equal(t, tc.expectedError.Error(), err.Error())
+			} else {
+				assert.NilError(t, tc.expectedError, err)
+			}
 		})
 	}
 }
--- a/provider/aci_test.go
+++ b/provider/aci_test.go
@@ -28,6 +28,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	testifyassert "github.com/stretchr/testify/assert"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 )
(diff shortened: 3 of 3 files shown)


helayoty:
> Don't you want to report the exact volume (v) that has the problem? Same on line 1649.

@eugen-nw I thought to handle that but decided to go with the same handling for AzureFile volume.

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			pv, err := p.resourceManager.GetPersistentVolume(v.Name)
+			if err != nil {
+				return volumes, err
```
ryanzhang-oss, line 1642: should we wrap the error here?

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+						ReadOnly:           csiSource.ReadOnly,
+						StorageAccountName: string(secret.Data["azurestorageaccountname"]),
+						StorageAccountKey:  string(secret.Data["azurestorageaccountkey"]),
+					},
+				}
+				volumes = append(volumes, volume)
+			}
+
+			continue
+		}
```
ryanzhang-oss, line 1667: I know the current code didn't do it but I think we should extract this into its own function.  We can refactor the rest piece by piece

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+			csiSource := pv.Spec.CSI
+
+			if csiSource.NodeStageSecretRef != nil && csiSource.NodeStageSecretRef.Name != "" {
+				secret, err := p.resourceManager.GetSecret(csiSource.NodeStageSecretRef.Name, csiSource.NodeStageSecretRef.Namespace)
+				if err != nil {
+					return volumes, err
+				}
+
+				if secret == nil {
+					return nil, fmt.Errorf("Getting secret for AzureFile CSI driver returned an empty secret")
```
ryanzhang-oss, line 1653: add the CSI source name?

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+				}
+				volume := aci.Volume{
+					Name: v.Name,
+					AzureFile: &aci.AzureFileVolume{
+						ReadOnly:           csiSource.ReadOnly,
+						StorageAccountName: string(secret.Data["azurestorageaccountname"]),
+						StorageAccountKey:  string(secret.Data["azurestorageaccountkey"]),
```
ryanzhang-oss, line 1660: where do these two constant string values come from?

provider/aci_test.go
```diff
@@ -1191,3 +1191,151 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
 		t.Fatal("Failed to create pod", err)
 	}
 }
+
+func TestCreatePodWithCSIVolume(t *testing.T) {
+	podName := "pod-" + uuid.New().String()
+	podNamespace := "ns-" + uuid.New().String()
```
ryanzhang-oss, line 1197: Why use a random string here? Do they have to be different?

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+			pv, err := p.resourceManager.GetPersistentVolume(v.Name)
+			if err != nil {
+				return volumes, err
+			}
+			csiSource := pv.Spec.CSI
+
+			if csiSource.NodeStageSecretRef != nil && csiSource.NodeStageSecretRef.Name != "" {
+				secret, err := p.resourceManager.GetSecret(csiSource.NodeStageSecretRef.Name, csiSource.NodeStageSecretRef.Namespace)
+				if err != nil {
+					return volumes, err
```
ryanzhang-oss, line 1649: wrap the error?

provider/aci.go
ryanzhang-oss: are they defined somewhere already?

provider/aci_test.go
```diff
@@ -1191,3 +1191,151 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+
+	mockCtrl := gomock.NewController(GinkgoT())
+	podLister := NewMockPodLister(mockCtrl)
+	secretLister := NewMockSecretLister(mockCtrl)
+	secretNamespaceLister := NewMockSecretNamespaceLister(mockCtrl)
+	configMapLister := NewMockConfigMapLister(mockCtrl)
+	serviceLister := NewMockServiceLister(mockCtrl)
+	pvcLister := NewMockPersistentVolumeClaimLister(mockCtrl)
+	pvLister := NewMockPersistentVolumeLister(mockCtrl)
+
```
ryanzhang-oss, line 1225: why are all the variables above called mockxxx while those are not called mock?

provider/aci_test.go
```diff
@@ -1191,3 +1191,151 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
 		t.Fatal("Failed to create pod", err)
 	}
 }
+
+func TestCreatePodWithCSIVolume(t *testing.T) {
```
ryanzhang-oss, line 1195: Please follow the table driven tests model (https://yourbasic.org/golang/table-driven-unit-test/)

provider/aci_test.go
```diff
@@ -1191,3 +1191,151 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+		assert.Check(t, is.Equal(fakeSubscription, subscription), "Subscription doesn't match")
+		assert.Check(t, is.Equal(fakeResourceGroup, resourceGroup), "Resource group doesn't match")
+		assert.Check(t, cg != nil, "Container group is nil")
+		assert.Check(t, is.Equal(podNamespace+"-"+podName, containerGroup), "Container group name is not expected")
+		assert.Check(t, cg.ContainerGroupProperties.Containers != nil, "Containers should not be nil")
+		assert.Check(t, is.Equal(1, len(cg.ContainerGroupProperties.Containers)), "1 Container is expected")
+		assert.Check(t, is.Equal("nginx", cg.ContainerGroupProperties.Containers[0].Name), "Container nginx is expected")
+		assert.Check(t, is.Equal(1, len(cg.Volumes)), "volume count not match")
+		assert.Check(t, is.Equal(pvName, cg.Volumes[0].Name), "volume name doesn't match")
+
```
ryanzhang-oss, line 1288: please follow the 

expected, actual style assert/compare. I have written extensive examples in the caravel repo

provider/aci_test.go
ryanzhang-oss: I have written extensive examples in the caravel repo

provider/aci_test.go
```diff
@@ -1191,3 +1191,151 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+				NodePublishSecretRef: &v1.LocalObjectReference{
+					Name: fakeVolumeSecret,
+				},
+			},
+		},
+	}
+	pod.Spec.Volumes = append(pod.Spec.Volumes, volume)
+
+	if err := provider.CreatePod(context.Background(), pod); err != nil {
+		t.Fatal("Failed to create pod", err)
```
ryanzhang-oss, line 1339: that's all we check on the error !=nil?

provider/aci_test.go
ryanzhang-oss: Please try to cover all the code in this PR.

provider/aci.go
```diff
@@ -1635,6 +1635,37 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			pv, err := p.resourceManager.GetPersistentVolume(v.Name)
+			if err != nil {
+				return volumes, err
+			}
+			csiSource := pv.Spec.CSI
+
+			if csiSource.NodeStageSecretRef != nil && csiSource.NodeStageSecretRef.Name != "" {
```
ryanzhang-oss, line 1646: what if csiSource.NodeStageSecretRef == nil? Is there any validator that will reject this beforehand?

helayoty: I wanted to ask if we should check on the `NodePublishSecretRef` property as well? how to know that? any documentation?

provider/aci.go
helayoty: ref from here : https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md

provider/aci.go
helayoty: After digging more into CSI for pods, only the `NodePublishSecretRef` needed to be checked. In case it's empty (as it's optional), we will not be able to add this volume to the ACI volume list .

ref: https://docs.microsoft.com/en-us/rest/api/container-instances/container-groups/create-or-update#volume

provider/aci_test.go
helayoty: I followed the code convention. I'll go ahead and follow the right convention for this case.

provider/aci_test.go
helayoty: Same. Followed the code convention. However, I'm ok with them being random.

provider/aci_test.go
helayoty: Same, followed the code convention. Happy to use expected/actual instead.

eugen-nw (commented):
Don’t you want to output the error as well?  It may save some head scratching.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.6.0
+	github.com/stretchr/testify v1.5.1
```
ryanzhang-oss, line 26: let's not add assert in this repo since it didn't use it

provider/aci.go
```diff
@@ -1632,9 +1635,54 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+func (p *ACIProvider) getAzureFileCSI(volume v1.Volume, namespace string) (*aci.Volume, error) {
+	azureSource := &aci.AzureFileVolume{
+		ReadOnly: *volume.CSI.ReadOnly,
+	}
+	if volume.CSI.NodePublishSecretRef != nil && volume.CSI.NodePublishSecretRef.Name != "" {
+
+		// Set the storage account name and key
+		secret, err := p.resourceManager.GetSecret(volume.CSI.NodePublishSecretRef.Name, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("The secret %s for AzureFile CSI driver %s is not found, err: %s", volume.CSI.NodePublishSecretRef.Name, volume.Name, err)
```
ryanzhang-oss, line 1647: please use errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found), volume.CSI.NodePublishSecretRef.Name)

in this way, the caller can still use errors.Is(err,targetErr) to handle different errors throw by the resourceManager

provider/aci.go
```diff
@@ -1632,9 +1635,54 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+}
+
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
+			if err != nil {
+				return volumes, err
```
ryanzhang-oss, line 1680: does it make sense to return a partial volume list?

provider/aci_test.go
```diff
@@ -1191,3 +1193,175 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "kube-root-ca.crt",
+		},
+		Data: map[string]string{
+			"ca.crt": "fake-ca-data",
+			"foo":    "bar",
+		},
+	}, nil)
+
+	mockSecretLister.EXPECT().Secrets(podNamespace).Return(mockSecretNamespaceLister)
```
ryanzhang-oss, line 1263: it seems that the test won't cover line 1647/1651 with this setting

provider/aci_test.go
```diff
@@ -1191,3 +1193,175 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+			},
+		},
+	}, nil)
+
+	readOnly := false
+
+	cases := []struct {
+		description   string
+		volume        v1.Volume
+		expectedError error
```
ryanzhang-oss, line 1294: is error the only thing to test? What about volume list?

provider/aci.go
helayoty: tbh, no but didn't know the impact of returning nil yet. This is how in tree files handled it.

provider/aci_test.go
helayoty: We had an assertion for the volume list length = 1

provider/aci.go
ryanzhang-oss: Please get some background on this, maybe Mitch knows. Andy should know too.

provider/aci.go
```diff
@@ -1632,9 +1635,54 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+		return nil, fmt.Errorf("NodePublishSecretRef for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
+	}
+}
+
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
```
ryanzhang-oss, line 1678: AFAIK, there are more Azure CSI than just file.  Should we check the v.CSI.Driver first?

helayoty: Yes, Azure Disk. I think it makes sense.

helayoty: Added a check for the case CSI driver `disk`, it will return error

provider/aci.go
```diff
@@ -1632,9 +1638,59 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+		return nil, fmt.Errorf("NodePublishSecretRef for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
+	}
+}
+
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			// Check if the CSI driver if file (Disk is not supported by ACI)
```
ryanzhang-oss, line 1681:
```suggestion
			// Check if the CSI driver is file (Disk is not supported by ACI)
```

provider/aci.go
```diff
@@ -1632,9 +1638,59 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+			if v.CSI.Driver == AzureFileDriverName {
+				csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
+				if err != nil {
+					return volumes, err
+				}
+				volumes = append(volumes, *csiVolume)
+				continue
+			} else {
+				return nil, fmt.Errorf("Pod %s requires volume %s which is of an unsupported type", pod.Name, v.Name)
```
ryanzhang-oss, line 1690:
```suggestion
				return nil, fmt.Errorf("Pod %s requires volume %s which is of an unsupported type %s", pod.Name, v.Name, v.CSI.Driver)
```

provider/aci.go
```diff
@@ -1632,9 +1638,59 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
 	volumes := make([]aci.Volume, 0, len(pod.Spec.Volumes))
 	for _, v := range pod.Spec.Volumes {
+		// Handle the case for Azure File CSI driver
+		if v.CSI != nil {
+			// Check if the CSI driver if file (Disk is not supported by ACI)
+			if v.CSI.Driver == AzureFileDriverName {
+				csiVolume, err := p.getAzureFileCSI(v, pod.Namespace)
+				if err != nil {
+					return volumes, err
```
ryanzhang-oss, line 1685: why return volumes here but not at line 1690?

provider/aci.go
helayoty: According to my conversation with Mitch; if there is an error, the volume will not be read by the codes. Check [here](https://github.com/virtual-kubelet/azure-aci/blob/14114f38278a90c7867ac4bf3ababb2fd5c33976/provider/aci.go#L672).

So, we can return `nil, err` instead of `volumes, err`

provider/aci.go
helayoty: fixed to be `nil`

provider/aci_test.go
```diff
@@ -1191,3 +1191,247 @@ func TestCreatePodWithProjectedVolume(t *testing.T) {
...
+			}
+			pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, volumeMount)
+
+			mockSecretLister.EXPECT().Secrets(podNamespace).Return(mockSecretNamespaceLister)
+
+			if tc.secretVolume == nil {
+				mockSecretNamespaceLister.EXPECT().Get(tc.volume.CSI.NodePublishSecretRef.Name).Return(nil, nil)
+			} else {
+				mockSecretNamespaceLister.EXPECT().Get(tc.secretVolume.Name).Return(tc.secretVolume, nil)
+			}
```
ryanzhang-oss, line 1424: I think this should be in the setup which mean the mockSecretNamespaceLister should be a parameter

helayoty: `mockSecretNamespaceLister` is not a parameter here, the return of the secret is the parameter (object or nil)

ryanzhang-oss (approved):

PR #199, opened 2022-05-06
bug: fix CSI driver inline implementation

helayoty:
Update the Azure file CSI driver inline volume support: https://docs.microsoft.com/en-us/azure/aks/azure-files-volume#mount-file-share-as-an-inline-volume

### DIFF
--- a/provider/aci.go
+++ b/provider/aci.go
@@ -50,7 +50,8 @@ const (
 	subnetDelegationService = "Microsoft.ContainerInstance/containerGroups"
 	// Parameter names defined in azure file CSI driver, refer to
 	// https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md
-	azureFileShareName = "shareName"
+	azureFileShareName  = "shareName"
+	azureFileSecretName = "secretName"
 	// AzureFileDriverName is the name of the CSI driver for Azure File
 	AzureFileDriverName = "file.csi.azure.com"
 )
@@ -1639,38 +1640,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
 }
 
 func (p *ACIProvider) getAzureFileCSI(volume v1.Volume, namespace string) (*aci.Volume, error) {
-	azureSource := &aci.AzureFileVolume{
-		ReadOnly: *volume.CSI.ReadOnly,
-	}
-	if volume.CSI.NodePublishSecretRef != nil && volume.CSI.NodePublishSecretRef.Name != "" {
+	azureSource := &aci.AzureFileVolume{}
 
-		// Set the storage account name and key
-		secret, err := p.resourceManager.GetSecret(volume.CSI.NodePublishSecretRef.Name, namespace)
-		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found", volume.CSI.NodePublishSecretRef.Name, volume.Name))
-		}
+	if volume.CSI.VolumeAttributes != nil && len(volume.CSI.VolumeAttributes) != 0 {
+		secretName := volume.CSI.VolumeAttributes[azureFileSecretName]
+		if secretName != "" {
+			secret, err := p.resourceManager.GetSecret(secretName, namespace)
 
-		if secret == nil {
-			return nil, fmt.Errorf("Getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
-		}
-
-		azureSource.StorageAccountName = string(secret.Data["azurestorageaccountname"])
-		azureSource.StorageAccountKey = string(secret.Data["azurestorageaccountkey"])
-
-		// Set shareName
-		if volume.CSI.VolumeAttributes != nil {
-			if shareName, ok := volume.CSI.VolumeAttributes[azureFileShareName]; ok {
-				azureSource.ShareName = shareName
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("the secret %s for AzureFile CSI driver %s is not found", secretName, volume.Name))
 			}
+
+			if secret == nil {
+				return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
+			}
+			// Set the storage account name and key
+
+			azureSource.StorageAccountName = string(secret.Data["azurestorageaccountname"])
+			azureSource.StorageAccountKey = string(secret.Data["azurestorageaccountkey"])
+		} else {
+			return nil, fmt.Errorf("secret name for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
 		}
-		volume := aci.Volume{
-			Name:      volume.Name,
-			AzureFile: azureSource,
+		// Set shareName
+
+		if shareName, ok := volume.CSI.VolumeAttributes[azureFileShareName]; ok {
+			azureSource.ShareName = shareName
 		}
-		return &volume, nil
 	} else {
-		return nil, fmt.Errorf("NodePublishSecretRef for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
+		return nil, fmt.Errorf("secret volume attribute for AzureFile CSI driver %s cannot be empty or nil", volume.Name)
 	}
+	return &aci.Volume{
+		Name:      volume.Name,
+		AzureFile: azureSource,
+	}, nil
 }
 
 func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]aci.Volume, error) {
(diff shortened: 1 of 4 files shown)


provider/aci.go
```diff
@@ -1639,38 +1640,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
-		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found", volume.CSI.NodePublishSecretRef.Name, volume.Name))
-		}
+	azureSource := &aci.AzureFileVolume{}
 
-		if secret == nil {
-			return nil, fmt.Errorf("Getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
-		}
+	if volume.CSI.VolumeAttributes != nil && len(volume.CSI.VolumeAttributes) != 0 {
+		secretName := volume.CSI.VolumeAttributes[azureFileSecretName]
```
andyzhangx, line 1646: field name in `CSI.VolumeAttributes` is case insensitive, you could refer to following code to handle such issue:

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/dabcf18dfa32fac089f45c182b63726726bc29b1/pkg/azurefile/azurefile.go#L568-L588

provider/aci.go
```diff
@@ -1639,38 +1640,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
-		// Set shareName
-		if volume.CSI.VolumeAttributes != nil {
-			if shareName, ok := volume.CSI.VolumeAttributes[azureFileShareName]; ok {
-				azureSource.ShareName = shareName
+			if secret == nil {
+				return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
 			}
+			// Set the storage account name and key
+
+			azureSource.StorageAccountName = string(secret.Data["azurestorageaccountname"])
```
andyzhangx, line 1659: use `strings.TrimSpace()`

provider/aci.go
```diff
@@ -1639,38 +1640,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+			azureSource.StorageAccountKey = string(secret.Data["azurestorageaccountkey"])
```
andyzhangx, line 1660: it's better define `azurestorageaccountkey` and `azurestorageaccountname` as const vars since they are referenced multiple times.

provider/aci.go
```diff
@@ -1639,38 +1642,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
-		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found", volume.CSI.NodePublishSecretRef.Name, volume.Name))
-		}
+	azureSource := &aci.AzureFileVolume{}
 
-		if secret == nil {
-			return nil, fmt.Errorf("Getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
-		}
+	if volume.CSI.VolumeAttributes != nil && len(volume.CSI.VolumeAttributes) != 0 {
+		secretName := volume.CSI.VolumeAttributes[azureFileSecretName]
```
andyzhangx, line 1648: it's case insensitive, not lower case, follow this code example to handle such case: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/dabcf18dfa32fac089f45c182b63726726bc29b1/pkg/azurefile/azurefile.go#L568-L588

provider/aci.go
```diff
@@ -1639,38 +1642,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
-		// Set shareName
-		if volume.CSI.VolumeAttributes != nil {
-			if shareName, ok := volume.CSI.VolumeAttributes[azureFileShareName]; ok {
-				azureSource.ShareName = shareName
+			if secret == nil {
+				return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
 			}
+			// Set the storage account name and key
+
+			azureSource.StorageAccountName = strings.TrimSpace(strings.ToLower(string(secret.Data[azureFileStorageAccountName])))
```
andyzhangx, line 1661: remove `strings.ToLower`

provider/aci.go
helayoty: Got it, please let me know how you think now.

mingqishao (dismissed):
look good for me.

provider/aci.go
```diff
@@ -1639,38 +1642,39 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
-	}
-	if volume.CSI.NodePublishSecretRef != nil && volume.CSI.NodePublishSecretRef.Name != "" {
+	azureSource := &aci.AzureFileVolume{}
 
-		// Set the storage account name and key
-		secret, err := p.resourceManager.GetSecret(volume.CSI.NodePublishSecretRef.Name, namespace)
-		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("The secret %s for AzureFile CSI driver %s is not found", volume.CSI.NodePublishSecretRef.Name, volume.Name))
-		}
+	if volume.CSI.VolumeAttributes != nil && len(volume.CSI.VolumeAttributes) != 0 {
```
andyzhangx, line 1647: I would suggest write like this, this would check field name first, also add a shareName filed check.
```
var secretName, shareName string
if volume.CSI.VolumeAttributes != nil {
	for k, v := range volume.CSI.VolumeAttributes {
		switch strings.ToLower(k) {
		case azureFileSecretName:
			secretName := v
		case azureFileShareName:
			shareName = v
	}
}

if shareName == "" {
	return error ...
}
azureSource.ShareName = shareName

if secretName == "" {
	return error ...
}

secret, err := p.resourceManager.GetSecret(secretName, namespace)
if err != nil {
	return nil, errors.Wrap(err, fmt.Sprintf("the secret %s for AzureFile CSI driver %s is not found", secretName, volume.Name))
}
if secret == nil {
	return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
}

// Set the storage account name and key
azureSource.StorageAccountName = strings.TrimSpace(string(secret.Data[azureFileStorageAccountName]))
azureSource.StorageAccountKey = strings.TrimSpace(string(secret.Data[azureFileStorageAccountKey]))
```

andyzhangx (dismissed):
/lgtm

provider/aci.go
```diff
@@ -1639,38 +1642,45 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("the secret %s for AzureFile CSI driver %s is not found", secretName, volume.Name))
 	}
+
+	if secret == nil {
+		return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
+	}
+
+	// Set the storage account name and key
+	azureSource.StorageAccountName = strings.TrimSpace(string(secret.Data[azureFileStorageAccountName]))
```
ryanzhang-oss, line 1678: what if the secret doesn't contain that azureFileStorageAccountName key?

helayoty: Good call. Added this corner case along with the unit tests

provider/aci.go
```diff
@@ -1639,38 +1642,49 @@ func getProbe(probe *v1.Probe, ports []v1.ContainerPort) (*aci.ContainerProbe, e
...
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("the secret %s for AzureFile CSI driver %s is not found", secretName, volume.Name))
+	}
+
+	if secret == nil {
+		return nil, fmt.Errorf("getting secret for AzureFile CSI driver %s returned an empty secret", volume.Name)
 	}
+
+	// Set the storage account name and key
+	if secret.Data[azureFileStorageAccountName] == nil || secret.Data[azureFileStorageAccountKey] == nil {
```
andyzhangx, line 1678: the `secret.Data[azureFileStorageAccountName]` actually returns `[]byte`, and `string([]byte{})` return empty string, so actually that is not a problem, no panic if no check.

`Data map[string][]byte`

helayoty: @andyzhangx So no change is required?

andyzhangx: yes, that check is not necessary, let's make code clean

helayoty: done

provider/aci.go
helayoty: @ryanzhang-oss  according to Andy's comment [here](https://github.com/virtual-kubelet/azure-aci/pull/199#discussion_r870896485), no need for this check.

ryanzhang-oss (approved):

PR #204, opened 2022-06-09
feat: InitContainer support for ACI Connector

t-ysalazar:
This change adds ACI initContainers virtual node support to virtual kubelet. The user can add ACI initContainers to the pod definition.

- Added InitContainer method and called on CreatePod so initContainers from the pod specs are created. 
- Factorize common methods between getInitContainer and getContainer.
- Throw errors for properties no supported on initContainers.
- Unit tests for InitContainers.
- E2E test for InitContainers.

### ACI InitContainerDefinition
According to ACI documentation, InitContainerDefinition unlike Container does not support livenessProbe, ports, readinessProbe and resources.

image

# Test

## Validated the change by testing on minikube.
image
image

### 1.
image
image
### 2.
image
image
### 3.
image
image

## Validated the change by testing on AKS.
image

image
image

# References
https://docs.microsoft.com/en-us/rest/api/container-instances/container-groups/create-or-update#request-body
https://docs.microsoft.com/en-us/rest/api/container-instances/container-groups/create-or-update#initcontainerdefinition
https://docs.microsoft.com/en-us/rest/api/container-instances/container-groups/create-or-update#container
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/


mainaksarcar (changes requested):

.gitignore
```diff
@@ -5,6 +5,9 @@
 *.so
 *.dylib
 
+helm-start.sh
+podspcs.yaml
```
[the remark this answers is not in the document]
t-ysalazar: Done

client/aci/types.go
```diff
@@ -72,6 +72,12 @@ type Container struct {
 	ContainerProperties `json:"properties,omitempty"`
 }
 
+// InitContainerDefinition is a initContainer instance.
+type InitContainerDefinition struct {
```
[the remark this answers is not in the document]
t-ysalazar: Done

mainaksarcar:
@helayoty - Can you please help review this PR for InitContainers support?

client/aci/types.go
```diff
@@ -65,6 +65,21 @@ type Container struct {
...
+
+// InitContainerProperties is the initContainer instance properties.
+type InitContainerProperties struct {
+	Image                string                          `json:"image,omitempty"`
+	Command              []string                        `json:"command,omitempty"`
+	EnvironmentVariables []EnvironmentVariable           `json:"environmentVariables,omitempty"`
+	InstanceView         ContainerPropertiesInstanceView `json:"instanceView,omitempty"`
+	VolumeMounts         []VolumeMount                   `json:"volumeMounts,omitempty"`
+}
+
```
helayoty, line 82: We don't need that anymore, you can use the SDK [models](https://github.com/Azure/azure-sdk-for-go/blob/main/services/containerinstance/mgmt/2021-10-01/containerinstance/models.go#L874) now.

provider/aci.go
```diff
@@ -680,6 +680,11 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 	containerGroup.RestartPolicy = aci.ContainerGroupRestartPolicy(pod.Spec.RestartPolicy)
 	containerGroup.ContainerGroupProperties.OsType = aci.OperatingSystemTypes(p.operatingSystem)
 
+	// get initContainers
```
helayoty, line 683: you don't have to add the functionality in the v1 code (before SDK). Please remove all new func

fnuarnav: removed from old code

client/aci/types.go
fnuarnav: removed

pkg/provider/aci.go
```diff
@@ -484,7 +484,15 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 		return err
 
 	}
+
+	// get initContainers
+	initContainers, err := p.getInitContainers(pod)
```
helayoty, line 489: Please pass the ctx on each new func and use it to log error

pkg/provider/aci.go
```diff
@@ -1213,6 +1221,85 @@ func readDockerConfigJSONSecret(secret *v1.Secret, ips []azaci.ImageRegistryCred
...
+func (p *ACIProvider) getInitContainers(pod *v1.Pod) ([]azaci.InitContainerDefinition, error) {
+	initContainers := make([]azaci.InitContainerDefinition, 0, len(pod.Spec.InitContainers))
+	for i, initContainer := range pod.Spec.InitContainers {
+		err := p.verifyContainer(&initContainer)
+		if err != nil {
+			return nil, err
+		}
+
+		if initContainer.Ports != nil {
+			return nil, errdefs.InvalidInput("ACI initContainers does not support ports.")
```
helayoty, line 1273: all error messages should start with lower case letter and no dots in the end.

pkg/provider/aci_init_container_test.go
```diff
@@ -0,0 +1,201 @@
...
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name: "container name",
+				},
+			},
+		},
+	}
+	cases := []struct {
```
helayoty, line 60: May we add test cases for init container with other containers. Also, it would be great if we add init container to one of the volumes unit tests.

README.md
```diff
@@ -36,6 +36,7 @@ Virtual Kubelet's ACI provider relies heavily on the feature set that Azure Cont
 * Basic Azure Networking support within AKS virtual node
 * [Exec support](https://docs.microsoft.com/azure/container-instances/container-instances-exec) for container instances
 * Azure Monitor integration or formally known as OMS
+* Support for init-containers ([use init containers](#Create-pod-with-init-containers))
```
helayoty, line 39: Can we update the limitations sections as well?

e2e/fixtures/initcontainers_pod.yml
```diff
@@ -0,0 +1,35 @@
...
+  initContainers:
+  - image: alpine
+    name: init-container-01
+    command: [ "/bin/sh" ]
+    args: [ "-c", "echo \"Hi\"" ]
+  - image: alpine
+    name: init-container-02
+    command: [ "/bin/sh" ]
+    args: [ "-c", "echo \"Hi\"" ]
+  containers:
```
helayoty, line 16: Can we add volume to the test here?

helayoty (changes requested):
few changes but overall almost there to merge it!

Fei-Guo:
l have a few questions: 

1. How to verify the initcontainer works as intended, i.e., it indeed runs BEFORE in the main container starts, in the e2e test? One idea is to let the main container and the init container share a volume and the init container writes a file and the main container reads the file. 

2. If Pod spec has invalid input, should we change the Pod status to indicate the error and avoid repeatedly reconcile the same Pod with wrong configuration? Note that the init container fields are immutable so it is ok for us stop reconciling the Pod with unsupported spec since the user has to recreate the Pod to fix the problem.

helayoty:
> l have a few questions:
> 
> 1. How to verify the initcontainer works as intended, i.e., it indeed runs BEFORE in the main container starts, in the e2e test? One idea is to let the main container and the init container share a volume and the init container writes a file and the main container reads the file.
> 2. If Pod spec has invalid input, should we change the Pod status to indicate the error and avoid repeatedly reconcile the same Pod with wrong configuration? Note that the init container fields are immutable so it is ok for us stop reconciling the Pod with unsupported spec since the user has to recreate the Pod to fix the problem.

 Very valid points @Fei-Guo, I'd love it if @t-ysalazar can apply the first point to the e2e test. However, the second point is a general topic that I'd like to discuss in depth with the team. I'll create an issue to track it now.

README.md
fnuarnav: removed init containers from limitations

pkg/provider/aci_init_container_test.go
fnuarnav: 2 containers are present in the Pod definition along with different types of initContainers, for all test cases

pkg/provider/aci.go
fnuarnav: updated error messages to use correct formatting

pkg/provider/aci.go
fnuarnav: logging error messges

helayoty (approved):
LGTM 🚢

mainaksarcar (dismissed):

e2e/pods_test.go
```diff
@@ -77,10 +78,6 @@ func TestPodLifecycle(t *testing.T) {
 	// check container logs
 	t.Log("get container logs ....")
 	cmd = kubectl("logs", "pod/vk-e2e-hpa", "-c", "hpa-example", "--namespace=vk-test")
-	out, err = cmd.CombinedOutput()
```
helayoty, line 80: Why did we remove lines 80-83

e2e/fixtures/initcontainers_ordertest_pod.yml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
```
helayoty, line 1: Can we combine the 2 pods/tests into one to save e2e test time? 
pod yaml file that contains:
- multiple initContainer
- InitContainer writes to a file and main container to verify that
- add volumes & volume mounts to the pod spec

e2e/pods_test.go
fnuarnav: this looks like a merge error. The merge conflicts were all over the place. I had moved the init container tests outside this file to avoid the same issue in the future.

pkg/provider/aci.go
```diff
@@ -840,6 +848,92 @@ func readDockerConfigJSONSecret(secret *v1.Secret, ips []azaci.ImageRegistryCred
...
+//this method is used for both initConainers and containers
+func (p *ACIProvider) getCommand(container *v1.Container) *[]string {
+	command := append(container.Command, container.Args...)
+	return &command
+}
+
+//get VolumeMounts declared on Container as []aci.VolumeMount
+func (p *ACIProvider) getVolumeMounts(container *v1.Container) *[]azaci.VolumeMount {
+	volumeMounts := make([]azaci.VolumeMount, 0, len(container.VolumeMounts))
+	fmt.Println(container.VolumeMounts)
```
helayoty, line 868: no fmt.println should be there, you can use log.G instead.

pkg/provider/aci.go
```diff
@@ -840,6 +848,92 @@ func readDockerConfigJSONSecret(secret *v1.Secret, ips []azaci.ImageRegistryCred
...
+			ReadOnly:  &container.VolumeMounts[i].ReadOnly,
+		})
+	}
+	return &volumeMounts
+}
+
+//get EnvironmentVariables declared on Container as []aci.EnvironmentVariable
+func (p *ACIProvider) getEnvironmentVariables(container *v1.Container) *[]azaci.EnvironmentVariable {
+	environmentVariable := make([]azaci.EnvironmentVariable, 0, len(container.Env))
+	for _, e := range container.Env {
```
helayoty, line 882: please use index instead of e to prevent duplicate cases.

pkg/provider/aci.go
fnuarnav: removed print stmt

pkg/provider/aci.go
fnuarnav: updated to use index

e2e/fixtures/initcontainers_ordertest_pod.yml
fnuarnav: using one combined e2e test for initcontainers

Fei-Guo (approved):

helayoty (approved):

PR #207, opened 2022-06-29
ci: Move to use GH actions

helayoty:
The current pipeline setup doesn't give access to contributors to access the result of pipelines running on the PRs. Decided to move to use Github actions instead and enhance the lint ci


mingqishao (approved):
Look good for meI'm not familiar with Github Action though, the configuration look good overall.

PR #244, opened 2022-08-10
Fnuarnav/delegated identity support

fnuarnav:
Delegated Identity support along with MI for image pull


client/aci/list.go
```diff
@@ -69,3 +68,69 @@ func (c *Client) ListContainerGroups(ctx context.Context, resourceGroup string)
 
 	return &list, nil
 }
+
+func (c *Client) ListAKSClusters(ctx context.Context, resourceGroup string) (*AKSClusterListResult, error) {
```
helayoty, line 72: After migrating to use nodeUtils, you should find an easier way to get k8s configs using 
`k8sClient, err := nodeutil.ClientsetFromEnv(kubeConfigPath)`

PR #278, opened 2022-08-31
chore: Use correct make deps for bin/% target

cpuguy83:
Now if any go files (minus e2e tests and test files) are modified, make
will rebuild.
Before this change you have to call `make clean build` to do a rebuild.


helayoty (approved):
lgtm

PR #279, opened 2022-08-31
ci: Use built-in duplicate detection

cpuguy83:
Instead of using the action job to detect duplicate runs, we can use
this `concurrency` option on the action to only run the given workflow
once for a particular commit.

Signed-off-by: Brian Goff <cpuguy83@gmail.com>


helayoty (approved):
lgtm

PR #299, opened 2022-09-26
chore: Use Azure SDK for auth

helayoty:
- Remove Azure German Cloud.
- Move the auth functionality into a separate folder and use Azure SDK instead.

old code path(s):
- https://github.com/virtual-kubelet/azure-aci/blob/master/provider/aci.go#L190-L260
- https://github.com/virtual-kubelet/azure-aci/blob/master/client/auth.go
- https://github.com/virtual-kubelet/azure-aci/blob/master/provider/acsCredential_test.go


pkg/auth/auth_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+const cred = `
+{
+    "cloud":"AzurePublicCloud",
+    "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47",
```
Fei-Guo, line 12: Is this a real tenantId? If both tenantId and subscriptionId are faked. please add a comment in the same line saying they are faked strings.

pkg/auth/auth.go
```diff
@@ -0,0 +1,264 @@
...
+
+		if subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID"); subscriptionID != "" {
+			c.AuthConfig.SubscriptionID = subscriptionID
+		}
+	}
+
+	resource := c.Cloud.Services[cloud.ResourceManager].Endpoint
+	options := &arm.ClientOptions{}
+
+	if cs, ok := options.Cloud.Services[cloud.ResourceManager]; ok {
```
Fei-Guo, line 142: In line 140, I think options are empty, so will line 142 be meaningless? You can defer the cleanup if this change only contains code move.

Fei-Guo (commented):
Please post the original code link in the change description, so we know this PR solely contains code moves. If not, please highlight the changed parts.

pkg/auth/auth.go
helayoty: For some reason, if we didn't add line 140, the `options.Cloud.Services[cloud.ResourceManager]` will return an empty string!

pkg/auth/auth_test.go
helayoty: The old code used to use a real tenant and subscription IDs! Planning to get rid of them in the next phase.

helayoty: Replaced.

pkg/auth/auth.go
helayoty: Removed and replaced by implicit import for arm package.

pkg/auth/auth.go
```diff
@@ -0,0 +1,250 @@
...
+		c.AuthConfig = auth
+	}
+
+	if aksCredFilepath := os.Getenv("AKS_CREDENTIAL_LOCATION"); aksCredFilepath != "" {
+		c.AKSCredential, err = NewAKSCredential(aksCredFilepath)
+		if err != nil {
+			return err
+		}
+
+		if c.AKSCredential != nil {
```
Fei-Guo, line 86: You may not need the nil check since if err == nil, the aksCredential should not be nil.

pkg/auth/auth.go
```diff
@@ -0,0 +1,250 @@
...
+	return nil
+}
+
+// Authentication represents the Authentication file for Azure.
+type Authentication struct {
+	ClientID             string `json:"clientId,omitempty"`
+	ClientSecret         string `json:"clientSecret,omitempty"`
+	SubscriptionID       string `json:"subscriptionId,omitempty"`
+	TenantID             string `json:"tenantId,omitempty"`
+	UseUserIdentity      bool   `json:"useUserIdentity,omitempty"`
```
Fei-Guo, line 148: UseUserIdentity flag can be derived from the length of the clientID field. You may not need this field in the struct.

pkg/auth/auth.go
```diff
@@ -0,0 +1,250 @@
...
+	UserIdentityClientId string `json:"userIdentityClientId,omitempty"`
+}
+
+// NewAuthenticationFromFile returns an Authentication struct from file path
+func (a *Authentication) NewAuthenticationFromFile(filepath string) error {
```
Fei-Guo, line 153: If you don't plan to expose this method to be called in other modules, make it an internal method (newAuthenticationFromFile).

pkg/auth/auth.go
```diff
@@ -0,0 +1,250 @@
...
+	ClientSecret           string `json:"aadClientSecret"`
+	ResourceGroup          string `json:"resourceGroup"`
+	Region                 string `json:"location"`
+	VNetName               string `json:"vnetName"`
+	VNetResourceGroup      string `json:"vnetResourceGroup"`
+	UserAssignedIdentityID string `json:"userAssignedIdentityID"`
+}
+
+// NewAKSCredential returns an aksCredential struct from file path
+func NewAKSCredential(p string) (*aksCredential, error) {
```
Fei-Guo, line 199: Can this be an internal method?  rename p-> filepath?

pkg/auth/auth.go
helayoty: The old design was taken into consideration if customers want to run the provider on unmanaged AKS.

pkg/auth/auth.go
helayoty: This is how the code determines the value [here](https://github.com/virtual-kubelet/azure-aci/pull/299/files/7563cac2f7a24ceafb6f8a6859dc0e610e01bb1a#diff-0ccbf9a2a7d709b87dbf2f30cc730528c6299d537bf721300d49502cfe521c0aR113).

Fei-Guo: I mean you don't need a field. You can expose a method
func (a Authentication) UseUserIdentity bool {
    return len(c.AuthConfig.ClientID) == 0
}
whenever you want to check this boolean.

pkg/auth/auth.go
Fei-Guo: I am checking the method "NewAKSCredential", &cred cannot be nil if err is nil.

pkg/auth/auth.go
helayoty: done

pkg/auth/auth.go
helayoty: done.

pkg/auth/auth.go
helayoty: done.

pkg/auth/auth.go
helayoty: done.

pkg/auth/auth.go
```diff
@@ -0,0 +1,250 @@
...
+			c.AuthConfig.ClientSecret = clientSecret
+		}
+
+		if userIdentityClientId := os.Getenv("VIRTUALNODE_USER_IDENTITY_CLIENTID"); userIdentityClientId != "" {
+			c.AuthConfig.UserIdentityClientId = userIdentityClientId
+		}
+
+		isUserIdentity := len(c.AuthConfig.ClientID) == 0
+
+		if isUserIdentity {
```
Fei-Guo, line 116: You can further remove "isUserIdentity" variable and add a comment instead. Feel free to do this in next change.

Fei-Guo (approved):

Issue #303, opened 2022-09-27
Virtual Kubelet Node is not capable of running pods throwing "DuplicateVolumes": Duplicate volumes 'kube-api-access-sndws' are found


tmcmm:
Since yesterday applicational pods stopped running throwing the following error:
```
Normal Scheduled 6m55s default-scheduler Successfully assigned xxxxx/test-e5ccbaf3-9xxxx5723ac2bc45-66654b7f99-f8t9c to virtual-node-aci-linux
Warning ProviderCreateFailed 75s (x16 over 6m55s) virtual-node-aci-linux/pod-controller api call to  https://management.azure.com/subscriptions/xxxxx/resourceGroups/MC_xxxxxx-aks-prod_westeurope/providers/Microsoft.ContainerInstance/containerGroups/xxxxxx?api-version=2021-07-01: got HTTP response status code 400 error code "DuplicateVolumes": Duplicate volumes 'kube-api-access-sndws' are found.
```

**Explanation:**
By default, a Pod is associated with a service account, and a credential (token) for that service account is placed into the filesystem tree of each container in that Pod, at /var/run/secrets/kubernetes.io/serviceaccount/token.
If available, a certificate bundle is placed into the filesystem tree of each container at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt, and should be used to verify the serving certificate of the API server.
Finally, the default namespace to be used for namespaced API operations is placed in a file at /var/run/secrets/kubernetes.io/serviceaccount/namespace in each container.

Tried to redeploy the application on the same namespace.
Tried to redeploy the application on a new namespace.
Tried to re-enable the virtual node addon.

From the logs of aci-connector:
kubectl logs -l app=aci-connector-linux -n kube-system
```
[...]

PR #305, opened 2022-09-27
fix: Fix multi-volume issue

helayoty:
- Fix volume issue #303 
- Add unit test to cover multiple volumes in case volume type is projected.
- Add e2e test case for pod with multi volume.

Root cause: Volume is not copied before it is appended to the list which causes all of the volumes in the list to have the same name because all of them refer to the same object in the range loop.


provider/aci.go
```diff
@@ -1720,10 +1720,10 @@ func (p *ACIProvider) getVolumes(pod *v1.Pod) ([]azaci.Volume, error) {
 			storageAccountKeyStr := string(secret.Data[azureFileStorageAccountKey])
 
 			volumes = append(volumes, azaci.Volume{
-				Name: &v.Name,
+				Name: &volume.Name,
```
Fei-Guo, line 1723: Do you need to copy volume name here as well?

helayoty: No need to copy the name here as we already copied the whole volume object.

Updated the unit and e2e tests for CSI drivers with multi-volumes.

Fei-Guo (approved):

PR #307, opened 2022-09-28
fix duplicate volume names error reported by container group service 

MartinForReal:
It is due to line 1738, volume is not copied before it is appended to the list. 
And all of the volumes in the list have the same name because all of them refer to the v in range loop.


helayoty:
duplicate with #305

MartinForReal:
duplicated with #305

PR #321, opened 2022-10-06
chore: Update new clients and add metrics client

helayoty:


pkg/client/client.go
```diff
@@ -6,15 +6,48 @@ import (
...
 
 const (
 	DefaultUserAgent = "virtual-kubelet/azure-arm-aci"
+	APIVersion       = "2021-10-01"
 )
 
 type ContainerGroupPropertiesWrapper struct {
-	azaci.ContainerGroupProperties
-	Extensions []*Extension `json:"client,omitempty"`
+	// ProvisioningState - READ-ONLY; The provisioning state of the container group. This only appears in the response.
```
helayoty, line 18: The API needs to check on properties and gives errors. Had to revert it back to copy all proprties.

pkg/client/client.go
```diff
@@ -48,15 +81,21 @@ func (c *ContainerGroupsClientWrapper) CreateCG(ctx context.Context, resourceGro
...
+				err = autorest.NewErrorWithError(err, "containerinstance.ContainerGroupsClient", "CreateOrUpdate", result.Response(), "Failure sending request trying again")
+				return err
+			}
+			return err
+		})
 	if err != nil {
-		err = autorest.NewErrorWithError(err, "containerinstance.ContainerGroupsClient", "CreateOrUpdate", result.Response(), "Failure sending request")
-		return err
-	}
-	if result.Response().StatusCode != http.StatusOK {
```
helayoty, line 56: The API is not consistent with the returned status core, it has multi values (200, 201, 0 or nil)

PR #353, opened 2022-11-02
fix: Fix duplicate container ports

helayoty:


Fei-Guo:
I think the problem happens when you have one container with multiple Ports, right? Should you add that test as well?

Fei-Guo (approved):

PR #356, opened 2022-11-04
fix: Duplicate Container port fix

helayoty:


pkg/provider/aci.go
```diff
@@ -499,11 +499,11 @@ func (p *ACIProvider) CreatePod(ctx context.Context, pod *v1.Pod) error {
 		count = count + len(*container.Ports)
 	}
 	ports := make([]azaci.Port, 0, count)
-	for _, container := range *containers {
-		for _, containerPort := range *container.Ports {
-
+	for c := range *containers {
+		container := ((*containers)[c]).Ports
```
Fei-Guo, line 503: The variable name `container` looks confusing. Can we rename it to `containerPorts`?

helayoty: done

Fei-Guo (approved):

PR #375, opened 2022-11-28
docs: Update DOWNGRADE-README.md

anson627:
Signed-off-by: Anson Qian <anson627@gmail.com>


helayoty (approved):

helayoty:
Thanks, @anson627 for your PR. While your change was valid, after having a conversation with @Fei-Guo, we can guarantee backward compatibility no forward compatibility.

PR #388, opened 2022-12-13
feat: Add a validation  init container

helayoty:
- Create Dockerfile and pkg folder for init container
- Add support for init container in helm chart, deployments for e2e tests
- Add support for init container in pipelines 

Signed-off-by: Heba Elayoty <hebaelayoty@gmail.com>


cmd/init-container/main.go
```diff
@@ -0,0 +1,46 @@
...
+			log.G(ctx).Fatal(err)
+		}
+	}
+	p := provider.ACIProvider{
+		ProviderNetwork: network.ProviderNetwork{},
+	}
+	// Check or set up a network for VK
+	err = p.ProviderNetwork.SetVNETConfig(ctx, &azConfig)
+	if err != nil {
+
```
Fei-Guo, line 44: What do we do for the error msg?  I think this is the key for the init container to provide sufficient log information for how to fixing errors and providing remediations.

helayoty: Added more tracing and logs.

cmd/init-container/main.go
```diff
@@ -0,0 +1,109 @@
...
+	config, err := clientcmd.BuildConfigFromFlags("", "")
+	if err != nil {
+		log.G(ctx).Fatal("an error has occurred while creating client ", err)
+	}
+
+	kubeClient := kubernetes.NewForConfigOrDie(config)
+	eventBroadcast := util.NewRecorder(ctx, kubeClient)
+	defer eventBroadcast.Shutdown()
+
+	recorder := eventBroadcast.NewRecorder(scheme.Scheme, v1.EventSource{Component: "k8s.io.api.core.v1.Pod"})
```
Fei-Guo, line 54: I think the component should be the agent name that fires the event, so it should be "virtual kubelet" or something similar.

cmd/init-container/main.go
```diff
@@ -0,0 +1,109 @@
...
+		})
+
+	if err != nil {
+		recorder.Eventf(&v1.ObjectReference{
+			Kind:      "Pod",
+			Name:      podName,
+			Namespace: podNamespace,
+		}, v1.EventTypeWarning, "InitFailed", "VNet config setup failed")
+		log.G(ctx).Fatal("cannot setup the VNet configuration ", err)
+	}
```
Fei-Guo, line 102: You should add a `return` at line 102.

pkg/util/events.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+func NewRecorder(ctx context.Context, kubeClient *kubernetes.Clientset) record.EventBroadcaster {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartStructuredLogging(3)
+	if eventBroadcaster != nil && kubeClient != nil {
```
Fei-Guo, line 18: Can eventBroadcaster be nil in this case?

pkg/network/aci_network.go
```diff
@@ -42,36 +43,49 @@ type ProviderNetwork struct {
 }
 
 func (pn *ProviderNetwork) SetVNETConfig(ctx context.Context, azConfig *auth.Config) error {
+	ctx, span := trace.StartSpan(ctx, "network.SetVNETConfig")
+	defer span.End()
+
+	if azConfig.AKSCredential != nil {
```
Fei-Guo, line 49: Is this a new behavior?

helayoty: No, this is only to add a trace for the 'SetVNETConfig' method.

cmd/init-container/main.go
helayoty: The Fatal will execute os.Exit(1) as implemented in the [logrus](https://github.com/sirupsen/logrus/blob/master/entry.go#L383) library.

pkg/util/events.go
helayoty: No need. Removed.

cmd/init-container/main.go
helayoty: Updated.

Fei-Guo (approved):

PR #391, opened 2022-12-14
refactor: Organize network module

helayoty:


pkg/network/aci_network.go
```diff
@@ -299,11 +283,15 @@ func formDNSSearchFitsLimits(ctx context.Context, searches []string) string {
 	return strings.Join(searches, " ")
 }
 
-func getProtocol(pro v1.Protocol) azaci.ContainerNetworkProtocol {
-	switch pro {
-	case v1.ProtocolUDP:
-		return azaci.ContainerNetworkProtocolUDP
-	default:
-		return azaci.ContainerNetworkProtocolTCP
+func omitDNSServerDuplicates(strs []string) []string {
```
Fei-Guo, line 286: Why do you introduce omitDNSServerDuplications? The method like the same as OmitDuplicates.

helayoty: To prevent import cycle between `provider` & `network` packages.

Issue #410, opened 2023-01-05
Update azure-aci to use nodeutil instead of node-cli


helayoty:
As part of ongoing efforts of upgrading the code base and removing unnecessary dependencies. We want to move away from the `node-cli` library and use `nodeutil` instead.

PR #277, opened 2022-08-30
chore: Update azure-aci to use nodeutil instead of node-cli

cpuguy83:
Fix webhook certificate auth with nodeutil
    
1. Routes were not setup and causing panics
2. The CA was not being loaded into the webhook auth and as such only
     token auth was available.

Issue #191, opened 2022-03-29
Upgrade node version to 1.21 ?


Socolin:
Hello,

In my AKS cluster all my nodes are upgraded to the 1.21 version but tue virtual nodes are still at 1.19, is there any plan to upgrade this to the latest version ?

![image](https://user-images.githubusercontent.com/539620/160645525-2fedd661-d3d9-45a8-895d-c0a08fe97889.png)

PR #411, opened 2023-01-05
refactor: Update azure-aci to use nodeutil instead of node-cli

helayoty:
- Move away from the `node-cli` library and use the built-in `nodeutil` package from virtual-kubelet instead.
- Use the new [virtual-kubelet](https://github.com/virtual-kubelet/virtual-kubelet/releases/tag/v1.8.0) version, `v1.8.0`.
- Upgrade Kubernetes version to `v1.26.0`
  - Update Liveness/ReadinessProbe (Prope -> PropeHandler)

Fixes: #410, #191

Update the old PR #277

Signed-off-by: Heba Elayoty <hebaelayoty@gmail.com>


pkg/provider/aci.go
```diff
@@ -756,10 +763,13 @@ func (p *ACIProvider) Ping(ctx context.Context) error {
 	return nil
 }
 
+func (p *ACIProvider) NotifyNodeStatus(ctx context.Context, cb func(*v1.Node)) {
```
cpuguy83, line 766: When VK creates a node it sets the node status to pending.
The node provider is resposible for telling when the node is ready.
Since the implementation here is a no-op, the node is never marked ready.

I think when creating the provider you can return a nil node provider and VK should use the default implementation which should work fine (or at least like before).

helayoty: Tried that, but it didn't solve the metrics issue.

cpuguy83 (approved):

go.mod
```diff
@@ -3,110 +3,127 @@ module github.com/virtual-kubelet/azure-aci
...
+	github.com/spf13/cobra v1.5.0
+	github.com/stretchr/testify v1.8.0
+	github.com/virtual-kubelet/virtual-kubelet v1.8.0
+	go.opencensus.io v0.23.0
+	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
 	gotest.tools v2.2.0+incompatible
-	k8s.io/api v0.21.0
-	k8s.io/apimachinery v0.21.0
-	k8s.io/client-go v0.21.0
+	k8s.io/api v0.26.2
```
Fei-Guo, line 29: Is api version 0.26.2 matching with client go version 0.25.0?

helayoty: It seems dependencies in VK have been updated to 0.26.0, updated here as well.

pkg/provider/aci_test.go
```diff
@@ -133,7 +133,7 @@ func TestCreatePodWithoutResourceSpec(t *testing.T) {
 
 		return nil
 	}
-	provider, err := createTestProvider(aciMocks, nil)
+	provider, err := createTestProvider(aciMocks, nil, nil, nil)
```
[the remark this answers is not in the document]
helayoty: Added initialized objects.

pkg/provider/podsTracker.go
```diff
@@ -173,6 +186,7 @@ func (pt *PodsTracker) processPodUpdates(ctx context.Context, pod *v1.Pod) bool
 
 	if err != nil {
 		log.G(ctx).WithError(err).Errorf("failed to retrieve pod %v status from provider", pod.Name)
+		return false
```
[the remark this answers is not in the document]
helayoty: Reverted

Fei-Guo (approved):

PR #476, opened 2023-03-01
fix: Use GetCG to retrieve InstanceView info

helayoty:
- The GetContainerGroupListResult API doesn't return the `InstanceView` status, which can cause nil. To mitigate this issue until ACI fixes it for VK, we had to get the CG info for each active CG. 
- Remove return in GetPods and replace it with logs + continue


pkg/provider/podsTracker.go
```diff
@@ -150,7 +150,8 @@ func (pt *PodsTracker) processPodUpdates(ctx context.Context, pod *v1.Pod) bool
 			pod.Status.Message = statusMessageNotFound
 			now := metav1.NewTime(time.Now())
 			for i := range pod.Status.ContainerStatuses {
-				if pod.Status.ContainerStatuses[i].State.Running == nil {
+				if pod.Status.ContainerStatuses != nil &&
```
Fei-Guo, line 153: Why do you need the nil check for ContainerStatuses? If it is nil, the range based for loop will be no-op. The program cannot reach here.

pkg/provider/aci.go
```diff
@@ -680,25 +680,50 @@ func (p *ACIProvider) GetPods(ctx context.Context) ([]*v1.Pod, error) {
...
 			continue
 		}
 
-		pod, err := p.containerGroupToPod(ctx, cgs[cgIndex])
-		if err != nil {
-			log.G(ctx).WithFields(log.Fields{
-				"name": cgs[cgIndex].Name,
-				"id":   cgs[cgIndex].ID,
-			}).WithError(err).Errorf("error converting container group %s to pod", cgs[cgIndex].Name)
+		if cg.Tags != nil && (*cg.Tags["NodeName"] != p.nodeName) {
```
Fei-Guo, line 711:
```go
if cg.Tags != nil && cg.Tags["NodeName"] != nil {
    if *cg.Tags["NodeName"] != p.nodeName {
     // Add a log
     continue
    }
} else {
   // Add a log, the cg is supposed to have a tag for nodename.
    continue
}
```

pkg/provider/podsTracker.go
helayoty: removed

pkg/provider/aci.go
helayoty: Done.

pkg/provider/aci.go
```diff
@@ -680,25 +680,63 @@ func (p *ACIProvider) GetPods(ctx context.Context) ([]*v1.Pod, error) {
...
 		}
 
-		pod, err := p.containerGroupToPod(ctx, cgs[cgIndex])
-		if err != nil {
+		if cg.Tags != nil && cg.Tags["NodeName"] != nil {
+			if *cg.Tags["NodeName"] != p.nodeName {
+				log.G(ctx).WithFields(log.Fields{
+					"name": *cgName,
+					"id":   *cg.ID,
+				}).Warnf("container group %s node name does match %s", *cgName, p.nodeName)
```
Fei-Guo, line 716: does match -> does not match

helayoty: Done

Fei-Guo (approved):

PR #479, opened 2023-03-06
feat: report both old and new styles of node OS information

allxiao:
> spec.template.spec.nodeSelector[beta.kubernetes.io/os]: deprecated since v1.14; use "kubernetes.io/os" instead

To maintain compatibility, the virtual node should declare both styles of OS information, so that the Pod being scheduled to the virtual node can choose any of the following options
* use the legacy style `beta.kubernetes.io/os` nodeSelector
* migrate to use the new `kubernetes.io/os` nodeSelector before the `beta` one is dropped completely from API server

Currently, we only report `beta.kubernetes.io/os`, which will block the users from working on the nodeSelector migration.

### DIFF
--- a/pkg/provider/vk_node.go
+++ b/pkg/provider/vk_node.go
@@ -7,6 +7,7 @@ package provider
 import (
 	"context"
 	"os"
+	"strings"
 
 	"github.com/virtual-kubelet/virtual-kubelet/trace"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Fei-Guo:
I'd rather go with just "beta.kubernetes.io/os", the current k8s API version of VK is still low.

allxiao:
Hi @Fei-Guo 

The `beta.kubernetes.io/os` label was deprecated since v1.14. Now the K8s API version for VK is 1.19. I think we should follow the deprecation process to add the support for the `kubernetes.io/os` label.

The labels on node defines the capabilities for a node. When we are in such kind of label deprecation, the node should declare it supports both labels, so that the users who want to schedule pods to VK can choose any one of the following:
1. use legacy `beta.kubernetes.io/os` nodeSelector
2. migrate to use `kubernetes.io/os` nodeSelector before the `beta` one is removed completely

When we only provide `beta.kubernetes.io/os`, the second one won't work, which blocks the user from preparing for the label deprecation and removal.

Fei-Guo:
Thanks.  I think the "beta.kubernetes.io/os" label already exists in vNode object. We just miss the ""kubernetes.io/os" label. 
It seems the node-cli does update the label 
https://github.com/virtual-kubelet/node-cli/blob/85167f75d00837452dbd6efe6e51512cdc65615e/internal/commands/root/node.go#L62

Since Heba is working on remove node cli dependency from azure-aci repo and I think she can add the new label in the change.
@helayoty.

helayoty:
@allxiao may I ask you to rebase your PR?

helayoty (approved):
lgtm

PR #484, opened 2023-03-13
fix: migrate from k8s.gcr.io to registry.k8s.io

ahmedwaleedmalik:
Kubernetes is migrating its image registry from [k8s.gcr.io](http://k8s.gcr.io/) to [registry.k8s.io](http://registry.k8s.io/).

Part of kubernetes/k8s.io#4780.

### DIFF
--- a/e2e/fixtures/csi-driver.yml
+++ b/e2e/fixtures/csi-driver.yml
@@ -6,7 +6,7 @@ metadata:
 spec:
   nodeName: vk-aci-test-aks
   containers:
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik:
> @ahmedwaleedmalik Thank you so much for your PR. Would you please update the rest of yaml test [files](https://github.com/virtual-kubelet/azure-aci/tree/master/e2e/fixtures) in the same folder as well?

Somehow I missed out on that, good catch.

helayoty (approved):

PR #486, opened 2023-03-16
chore: Upgrade go to 1.19

helayoty:
Move from golang 1.18 to 1.19


Fei-Guo:
This is needed to upgrade k8s api version to 1.26.

PR #496, opened 2023-03-23
test: unit test for makeRegistryCredentialFromDockerConfig

smritidahal653:
Added unit test for makeRegistryCredentialFromDockerConfig with 4 test cases. This unit test will increase the overall coverage for aci.go by 1.1%.


pkg/provider/aci_test.go
```diff
@@ -113,6 +114,64 @@ func TestMakeRegistryCredential(t *testing.T) {
...
+	password := "pass-" + uuid.New().String()
+	authConfig := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", username, password)))
+
+	tt := []struct {
+		name        string
+		authConfig  dockercfg.AuthConfig
+		shouldFail  bool
+		failMessage string
+	}{
+		{
```
Fei-Guo, line 130: What is the difference between the 1st and 2nd cases? They look the same.

pkg/provider/aci_test.go
```diff
@@ -113,6 +114,64 @@ func TestMakeRegistryCredential(t *testing.T) {
...
+			"Valid username and password",
+			dockercfg.AuthConfig{Username: username, Password: password},
+			false,
+			"",
+		},
+		{
+			"Username and password in auth",
```
Fei-Guo, line 137: Can you name the test more precisely? The sentence is hard to understand.

smritidahal653: updated name

Fei-Guo (approved):

PR #497, opened 2023-03-24
test: unit test for filterWindowsServiceAccountSecretVolume

smritidahal653:
Added unit test for filterWindowsServiceAccountSecretVolume. Also created unit test TestCreatePodWithWindowsOs which calls CreatePod method with the os set to "Windows"


pkg/provider/aci_test.go
```diff
@@ -687,10 +737,15 @@ func createTestProvider(aciMocks *MockACIProvider, configMapMocker *MockConfigMa
 
 	cfg.Node = &v1.Node{}
 
+	operatingSystem := os.Getenv("PROVIDER_OPERATING_SYSTEM")
```
helayoty, line 740: We can use https://pkg.go.dev/os@go1.19.1#LookupEnv instead.

pkg/provider/aci_test.go
```diff
@@ -1105,3 +1160,106 @@ func TestGetPodWithContainerID(t *testing.T) {
...
+			if tc.shouldFilter {
+				assert.Check(t, is.Equal(0, len(cg.Properties.Containers[0].Properties.VolumeMounts)), "should filter out volume mounts with service account secret volume name")
+				assert.Check(t, is.Equal(1, len(cg.Properties.Volumes)), "should filter out volume with service account secret volume name")
+			} else {
+				assert.Check(t, is.Equal(1, len(cg.Properties.Containers[0].Properties.VolumeMounts)), "volume mount should remain the same")
+				assert.Check(t, is.Equal(2, len(cg.Properties.Volumes)), "volume should remain the same")
+			}
+		})
+	}
+}
```
helayoty, line 1265: nit: add new line.

pkg/provider/aci_test.go
```diff
@@ -1105,3 +1160,106 @@ func TestGetPodWithContainerID(t *testing.T) {
...
+	cgName := "pod-" + uuid.New().String()
+	cgNamespace := "ns-" + uuid.New().String()
+	mockCtrl := gomock.NewController(t)
+ 	defer mockCtrl.Finish()
+
+	volName:= "fakeVolume"
+	volMountName1:= "fakeVolumeMount1"
+	volMountPath1:= "/mnt/azure"
+	volMountName2:= "fakeVolumeMount2"
+	serviceAccountSecretMountPath:= "/var/run/secrets/kubernetes.io/serviceaccount"
```
helayoty, line 1174: You can use this [const](https://github.com/virtual-kubelet/azure-aci/blob/master/pkg/provider/aci.go#L44) instead

helayoty (changes requested):
Few nits but overall really LGTM.

pkg/provider/aci_test.go
smritidahal653: The mount path requires a pointer, so I decided to stick with this

PR #512, opened 2023-04-03
test: added unit test for getImagePullSecrets with docker config JSON

smritidahal653:
added unit test for getImagePullSecrets in aci.go with dockerc config JSON secret.

### DIFF
--- a/pkg/provider/aci_test.go
+++ b/pkg/provider/aci_test.go
@@ -1494,18 +1494,12 @@ func TestGetImagePullSecretsWithDockerCfgSecret(t *testing.T) {
 	}
 
 	cases := []struct {
-		description      string
-		imagePullSecrets []v1.LocalObjectReference
-		callSecretMocks  func(secretMock *MockSecretLister)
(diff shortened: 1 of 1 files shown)


pkg/provider/aci_test.go
```diff
@@ -1582,3 +1582,168 @@ func TestGetImagePullSecretsWithDockerCfgSecret(t *testing.T) {
...
+		},
+		Type: v1.SecretTypeDockerConfigJson,
+		Data: map[string][]byte{
+			v1.DockerConfigJsonKey: []byte(validCfgJson),
+		},
+	}
+
+	cases := []struct {
+		description      string
+		imagePullSecrets []v1.LocalObjectReference
```
Fei-Guo, line 1649: I see imagePullSecrets is the same for all test cases. You can remove this variable and make it a const used in line 1727.

Fei-Guo (commented):
LGTM

helayoty (approved):

PR #516, opened 2023-04-05
test: added unit test for cleanupDanglingPods

smritidahal653:
added unit test for cleanupDanglingPods in podsTracker.go
- test case: podLister returns an additional pod than what is reported by the client so the dangling pod is deleted so the two are in sync


helayoty (approved):
lgtm 🚢

Fei-Guo (approved):

Issue #517, opened 2023-04-05
Add unit tests for pkg/auth/auth.go


nidhunge:
- [x] Add unit test for decode function #518  
- [ ] Pass ConfigInterface to SetAuthConfig method for  mock #533

PR #518, opened 2023-04-05
test: Added unit test for decode function in auth.go

nidhunge:
Issue Link: https://github.com/virtual-kubelet/azure-aci/issues/517


helayoty (approved):
lgtm

PR #520, opened 2023-04-06
test: Display code coverage for each func/pkg and the total coverage

helayoty:
Because the third-party extension (aka `codecov`) coverage can get delayed reporting back the code coverage, we wanted to display the full coverage report with the total with each run using go native way `go tool cover -func=coverage.out`.


helayoty:
Current actual code coverage:

https://github.com/virtual-kubelet/azure-aci/actions/runs/4631373829/jobs/8194154942?pr=520#step:4:574

Fei-Guo (approved):

PR #521, opened 2023-04-06
test: added unit test for updatePodsLoop

smritidahal653:
added unit test for updatePodsLoop in podsTracker.go

### DIFF
--- a/pkg/provider/podsTracker_test.go
+++ b/pkg/provider/podsTracker_test.go
@@ -268,17 +268,16 @@ func TestUpdatePodsLoop(t *testing.T) {
 	podName1 := "pod-" + uuid.New().String()
 	podName2 := "pod-" + uuid.New().String()
 	podName3 := "pod-" + uuid.New().String()
+	podName4 := "pod-" + uuid.New().String()
 	podNamespace := "ns-" + uuid.New().String()
 
-	podsNames := []string{podName1, podName2, podName3}
+	podsNames := []string{podName1, podName2, podName3, podName4}
 	k8sPods := testsutil.CreatePodsList(podsNames, podNamespace)
 
 	k8sPods[0].Status.Phase = v1.PodPending
 	k8sPods[1].Status.Phase = v1.PodRunning
-	k8sPods[2].Status.Phase = v1.PodPending
-
-	podsCopy := make([]*v1.Pod, len(k8sPods))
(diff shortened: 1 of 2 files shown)


pkg/provider/podsTracker_test.go
```diff
@@ -263,3 +263,81 @@ func TestCleanupDanglingPods(t *testing.T) {
...
+
+	if equality.Semantic.DeepEqual(k8sPods[0], podsCopy[0]) {
+		t.Errorf("pod has been updated so it should not be deeply equal to the original copy")
+	}
+	if equality.Semantic.DeepEqual(k8sPods[1], podsCopy[1]) {
+		t.Errorf("pod has been updated so it should not be deeply equal to the original copy")
+	}
+	if !equality.Semantic.DeepEqual(k8sPods[2], podsCopy[2]) {
+		t.Errorf("pod has not been updated, so it should remain the same as the original copy")
+	}
```
helayoty, line 332: We can have this as a for loop

pkg/provider/podsTracker_test.go
```diff
@@ -263,3 +263,81 @@ func TestCleanupDanglingPods(t *testing.T) {
...
+	podName2 := "pod-" + uuid.New().String()
+	podName3 := "pod-" + uuid.New().String()
+	podNamespace := "ns-" + uuid.New().String()
+
+	podsNames := []string{podName1, podName2, podName3}
+	k8sPods := testsutil.CreatePodsList(podsNames, podNamespace)
+
+	k8sPods[0].Status.Phase = v1.PodPending
+	k8sPods[1].Status.Phase = v1.PodRunning
+	k8sPods[2].Status.Phase = v1.PodPending
```
helayoty, line 278: May we add the `PodSucceeded ` and `PodFailed ` status phases as well?

smritidahal653: There is a separate unit test for processPodUpdates that tests all the statuses, do we need that in the current unit test as well?

helayoty: Ah, you're right. But I can see the `PodFailed` is covered, but `PodSucceeded` is not. Feel free to add it to the other UT.

Fei-Guo:
LGTM

helayoty (approved):

Issue #514, opened 2023-04-04
Add unit tests for real_time.go


nidhunge:
There are no unit tests for real_time.go file. 

- [x] Create and add real_time_test.go file #515
- [x]  Add Unit Test for calculateUsageNanoCores #515
- [x] Bugfix for edge scenarios when uint64 values used are negative or very low in real_time.go #515
- [x] Bugfix for pointer reference where the pointer jumped to next object even if the "If" condition wasn't valid in a for loop. #515
- [x] Add Unit Test for filterOutContainerNotInPod #526 
- [x] Add Unit Test for getRealTimeExtensionPodStats #531 
- [x] Add Unit Test for extensionPodStatsToKubeletPodStats #534 
- [x] Bugfix for extensionPodStatsToKubeletPodStats where pointers jumped inside for range loops and assigned the same value as the last item in range for all values.#534

PR #526, opened 2023-04-10
test: Added unit test for filterOutContainerNotInPod

nidhunge:
Issue Link: https://github.com/virtual-kubelet/azure-aci/issues/514

### DIFF
--- a/pkg/metrics/real_time_test.go
+++ b/pkg/metrics/real_time_test.go
@@ -204,7 +204,7 @@ func TestFilterOutContainerNotInPod(t *testing.T) {
 		},
 	}
 
-	testDescription := "Sucessfully filters out containerStats for containers not in pod.spec.containers"
+	testDescription := "Successfully filters out containerStats for containers not in pod.spec.containers"
 	t.Run(testDescription, func(t *testing.T) {
 		filterOutContainerNotInPod(podStats, pod)
 		assert.EqualValues(t, len(podStats.Containers), 2)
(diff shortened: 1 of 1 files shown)


pkg/metrics/real_time_test.go
```diff
@@ -148,3 +151,65 @@ func TestCalculateUsageNanoCores(t *testing.T) {
...
+					UsageCoreNanoSeconds: 2345678900,
+				},
+			},
+			{
+				Name: fake_container2,
+				CPU: cpuStats{
+					UsageCoreNanoSeconds: 1234567800,
+				},
+			},
+			{
```
helayoty, line 198: While the test case will pass but this will confuse the developer. As the realtime will get the metrics for all containers in this pod. It's impossible to have one pod runs different types of containers (k8s vs ACI).

nidhunge: Understood. Added fake-container4 realtime metrics in test

helayoty: But `fake_container3` is not in the containers list

nidhunge: Not sure I understand. Isn't the purpose of the function to filter out fake_container3 from containerStats as the container is not in the container list of pods?

nidhunge: https://github.com/virtual-kubelet/azure-aci/blob/c88ad2ee2db41b678120da70253c5162a0c5dc54/pkg/metrics/real_time.go#L268

helayoty: It's the other way, the realtime extension can retrieve other containers (side cars) which are not in the pod.

helayoty (approved):

PR #537, opened 2023-04-15
test: added unit test for validateNetworkConfig

smritidahal653:
added unit test coverage for validateNetworkConfig in aci_network.go

Test case break down:

### Vnet name
- Provider Vnet name is set but environment variable is not: tc2
- Provider Vnet name is not set but environment variable is: tc3
- Both provider Vnet name and environment variable is set: tc6
- Neither provider Vnet name nor environment variable is set: tc1

### Vnet resource group
- Provider Vnet resource group is set but environment variable is not: tc2
- Provider Vnet resource group is not set but environment variable is: tc4
- Both provider Vnet resource group and environment variable is set: tc6
- Neither provider Vnet resource group nor environment variable is set: tc3

### Subnet name
- Provider Subnet name is set but environment variable is not: tc2
- Provider Subnet name is not set but environment variable is
-- malformed: tc7
--valid: tc8
- Both provider Subnet name and environment variable is set: tc6
- Neither provider Subnet name nor environment variable is set: tc5

### DIFF
--- a/pkg/network/aci_network_test.go
+++ b/pkg/network/aci_network_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"testing"
 
+	aznetworkv2 "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	testsutil "github.com/virtual-kubelet/azure-aci/pkg/tests"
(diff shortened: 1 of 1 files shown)


pkg/network/aci_network_test.go
```diff
@@ -254,3 +257,83 @@ func TestShouldCreateSubnet(t *testing.T) {
...
 }
+
+func TestValidateNetworkConfig(t *testing.T) {
+	azConfig := auth.Config{}
+	err := azConfig.SetAuthConfig(context.TODO())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pn := &ProviderNetwork{}
```
helayoty, line 268: We need to move the pn inside the for [loop](https://github.com/virtual-kubelet/azure-aci/pull/537/files#diff-b6968d67da7e5445402e60e9f7865636274e1d7c85ad4c3b90276e4de34f27bfR323) so we are sure it's getting reset after each case.

pkg/network/aci_network_test.go
```diff
@@ -254,3 +257,83 @@ func TestShouldCreateSubnet(t *testing.T) {
...
+
+	cases := []struct {
+		description   string
+		setEnvVar     func()
```
helayoty, line 270: An important test case is what will happen if pn object has one or more property sets. for example, in the following piece of code

```go
if vnetResourceGroup := os.Getenv("ACI_VNET_RESOURCE_GROUP"); vnetResourceGroup != "" {
		log.G(ctx).Debug("ACI VNet resource group env variable ACI_VNET_RESOURCE_GROUP is set")

		pn.VnetResourceGroup = vnetResourceGroup
	} else if pn.VnetResourceGroup == "" {
		return errors.New("vnet resourceGroup can not be empty please set ACI_VNET_RESOURCE_GROUP")
	}
```
We need to check the following cases:
	- `ACI_VNET_RESOURCE_GROUP ` env var is set but`VnetResourceGroup` is not set.
	- `ACI_VNET_RESOURCE_GROUP ` env var is NOT set but`VnetResourceGroup` is set.
	-  Neither of them is set.
	- Both are set.
	
We may need to add the pn obj in the test case struct and test the different behaviors.

pkg/network/aci_network_test.go
smritidahal653: done

helayoty (approved):
Thanks for redesigning the test case. It hits all the corner cases now. LGTM 🚀

PR #547, opened 2023-05-23
fix: container state.

hypnoce:
started is always true. make pod with startup probe fail.
ensure terminated container group report a terminated state in k8s.

### DIFF
--- a/pkg/provider/containergroup_to_pod.go
+++ b/pkg/provider/containergroup_to_pod.go
@@ -65,12 +65,14 @@ func (p *ACIProvider) getPodStatusFromContainerGroup(ctx context.Context, cg *az
 			firstContainerStartTime = *containersList[0].Properties.InstanceView.CurrentState.StartTime
 			lastUpdateTime = firstContainerStartTime
 		}
-
+		containerState := aciContainerStateToContainerState(containersList[i].Properties.InstanceView.CurrentState)
+		started := containerState.Running != nil
 		containerStatus := v1.ContainerStatus{
 			Name:                 *containersList[i].Name,
-			State:                aciContainerStateToContainerState(containersList[i].Properties.InstanceView.CurrentState),
+			State:                containerState,
 			LastTerminationState: aciContainerStateToContainerState(containersList[i].Properties.InstanceView.PreviousState),
 			Ready:                getPodPhaseFromACIState(*containersList[i].Properties.InstanceView.CurrentState.State) == v1.PodRunning,
+			Started:              &started,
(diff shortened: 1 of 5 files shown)


pkg/provider/containergroup_to_pod.go
```diff
@@ -131,7 +132,7 @@ func aciContainerStateToContainerState(cs *azaciv2.ContainerState) v1.ContainerS
 			},
 		}
 	// Handle the case of completion.
-	case "Succeeded":
+	case "Succeeded", "Terminated":
```
mainaksarcar, line 135: The return should have a Reason other than "Completed" that goes with the "Terminated" state.

mainaksarcar (approved):

pkg/provider/containergroup_to_pod.go
hypnoce: Usually, CRIs set Reason to Completed when state is terminated: 
https://github.com/Mirantis/cri-dockerd/blob/39a4ff0619d49651055509a0d2c3895f33772329/core/container_status.go#LL92C15-L92C24
https://github.com/containerd/containerd/blob/4b7145cfd38d76bc763e3cee1210963bc6c22680/pkg/cri/server/container_status.go#LL91C31-L91C31

Maybe Terminated as another meaning in ACI ? What reason should it be ?

pkg/provider/containergroup_to_pod.go
```diff
@@ -65,12 +65,13 @@ func (p *ACIProvider) getPodStatusFromContainerGroup(ctx context.Context, cg *az
...
 			lastUpdateTime = firstContainerStartTime
 		}
-
+		started := true
 		containerStatus := v1.ContainerStatus{
 			Name:                 *containersList[i].Name,
 			State:                aciContainerStateToContainerState(containersList[i].Properties.InstanceView.CurrentState),
 			LastTerminationState: aciContainerStateToContainerState(containersList[i].Properties.InstanceView.PreviousState),
 			Ready:                getPodPhaseFromACIState(*containersList[i].Properties.InstanceView.CurrentState.State) == v1.PodRunning,
+			Started:              &started,
```
helayoty, line 74: What is the need to set this property here?

helayoty, line 74: How do we know that the container has started in ACI? Why do we assume it will be true?

hypnoce: In absence of startup  probe (not supported by ACI), the default value is true. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#containerstatus-v1-core

helayoty (approved):

PR #565, opened 2023-07-24
docs: Update the upgrade script

MahatiC:
Update the correct url for helm chart


MahatiC:
@helayoty Could you help me understand the reason for closing this? When I was trying to upgrade, one of the fixed variables in the README needed to be changed - so I raised this PR. Has it since been corrected?

PR #570, opened 2023-09-14
test: changing aks version to default

smritidahal653:
Removed hardcoded AKS version 1.24.10 which is no longer supported


helayoty (approved):
LGTM

PR #594, opened 2024-01-22
chore: updating permissions for workflow files

smritidahal653:
Updating/adding permissions block for workflow file according to [breaking change doc](https://docs.opensource.microsoft.com/github/apps/permission-changes/)

- List of available scopes: https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#permissions
- Permission breakdown per scope: https://docs.github.com/en/enterprise-cloud@latest/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,11 @@
 version: 2
+
+permissions:
+  pull-requests: write
+  security-events: write
+  contents: read
+  actions: read
+
 updates:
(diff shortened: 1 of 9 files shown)


helayoty (approved):
I think this good for now, we might need to watch the pipelines once the change rolled out to make sure all pipeline.

PR #628, opened 2024-07-31
fix: separate subnet create vs update and check for addressPrefixes field

smritidahal653:
- Added a check for subnet properties.AddressPrefixes incase properties.AddressPrefix is not set and added necessary unit tests
- To differentiate between subnet create vs update scenario:
  - changed the existing "createSubnet" boolean to "updateSubnet" to minimize confusion - this bool indicates that an update is needed
  - introduced new boolean "createNewSubnet" which will only be set to true if we cannot retrieve the ACI subnet from GetACISubnet function
  - Updated function "CreateACISubnet" to "CreateOrUpdateACISubnet" and distinguished create vs update scenario using the "createNewSubnet" boolean from above
  - updated logic so addressPrefix/addressPrefixes is only set on create

### DIFF
--- a/pkg/network/aci_network.go
+++ b/pkg/network/aci_network.go
@@ -137,9 +137,8 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
 	if currentSubnet == (aznetworkv2.Subnet{}) {
 		createNewSubnet = true
 	} else {
-		updateSubnet = true
 		// check if the current subnet is valid or if we need to create a new subnet
-		updateSubnet, err = pn.shouldCreateSubnet(currentSubnet, updateSubnet)
+		updateSubnet, err = pn.shouldCreateSubnet(currentSubnet)
 		if err != nil {
 			return err
 		}
@@ -157,7 +156,8 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
 	return nil
 }
 
-func (pn *ProviderNetwork) shouldCreateSubnet(currentSubnet aznetworkv2.Subnet, createSubnet bool) (bool, error) {
+func (pn *ProviderNetwork) shouldCreateSubnet(currentSubnet aznetworkv2.Subnet) (bool, error) {
+	createSubnet := true
 	//check if addressPrefix has been set
 	if currentSubnet.Properties.AddressPrefix != nil && len(*currentSubnet.Properties.AddressPrefix) > 0 {
 		if pn.SubnetCIDR == "" {
--- a/pkg/network/aci_network.go
+++ b/pkg/network/aci_network.go
@@ -179,10 +179,7 @@ func (pn *ProviderNetwork) isCurSubnetValid(currentSubnet aznetworkv2.Subnet, is
 	if currentSubnet.Properties.ServiceAssociationLinks != nil {
 		for _, l := range currentSubnet.Properties.ServiceAssociationLinks {
 			if l.Properties != nil && l.Properties.LinkedResourceType != nil {
-				if *l.Properties.LinkedResourceType == subnetDelegationService {
-					isValidSubnet = false
-					break
-				} else {
+				if *l.Properties.LinkedResourceType != subnetDelegationService {
 					return false, fmt.Errorf("unable to delegate subnet '%s' to Azure Container Instance as it is used by other Azure resource: '%v'", pn.SubnetName, l)
 				}
 			}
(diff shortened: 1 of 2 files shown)


pkg/network/aci_network.go
```diff
@@ -155,36 +158,48 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
...
 		if pn.SubnetCIDR == "" {
 			pn.SubnetCIDR = *currentSubnet.Properties.AddressPrefix
 		}
 		if pn.SubnetCIDR != *currentSubnet.Properties.AddressPrefix {
 			return false, fmt.Errorf("found subnet '%s' using different CIDR: '%s'. desired: '%s'", pn.SubnetName, *currentSubnet.Properties.AddressPrefix, pn.SubnetCIDR)
 		}
-		if currentSubnet.Properties.RouteTable != nil {
-			return false, fmt.Errorf("unable to delegate subnet '%s' to Azure Container Instance since it references the route table '%s'", pn.SubnetName, *currentSubnet.Properties.RouteTable.ID)
+	} else if len(currentSubnet.Properties.AddressPrefixes) > 0 { // else check if addressPrefixes array has been set
+		firstPrefix := currentSubnet.Properties.AddressPrefixes[0]
```
Fei-Guo, line 170: You need to add a nil check for the firsPreview.

pkg/network/aci_network.go
```diff
@@ -155,36 +158,48 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
...
-			}
-		} else {
-			for _, d := range currentSubnet.Properties.Delegations {
-				if d.Properties != nil && d.Properties.ServiceName != nil &&
-					*d.Properties.ServiceName == subnetDelegationService {
+		if pn.SubnetCIDR != *firstPrefix {
+			return false, fmt.Errorf("found subnet '%s' using different CIDR: '%s'. desired: '%s'", pn.SubnetName, *firstPrefix, pn.SubnetCIDR)
+		}
+	} else {
+		return false, fmt.Errorf("both AddressPrefix and AddressPrefixes for subnet '%s' are not set", pn.SubnetName)
```
Fei-Guo, line 178: There are other conditions where the fields are set but the string is an empty string. Try to make the error msg more accurate.

pkg/network/aci_network.go
```diff
@@ -128,23 +128,26 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
...
-	createSubnet := true
+	updateSubnet := true
 	currentSubnet, err := pn.GetACISubnet(ctxWithResp, subnetsClient)
 	if err != nil {
 		return err
 	}
 
-	if err == nil {
-		createSubnet, err = pn.shouldCreateSubnet(currentSubnet, createSubnet)
+	createNewSubnet := false
```
Fei-Guo, line 137: The code logic is correct. But can you slightly improve the code readability?

You can initialize updateSubnet to false.

line 148, change it to

if createNewSubnet || updateSubnet:

pkg/network/aci_network.go
```diff
@@ -128,23 +128,27 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
...
+	if currentSubnet == (aznetworkv2.Subnet{}) {
+		createNewSubnet = true
+	} else {
+		updateSubnet = true
```
Fei-Guo, line 141: You don't need this line, I believe. line 143 should give a correct value.

pkg/network/aci_network.go
```diff
@@ -128,23 +128,27 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
 	var rawResponse *http.Response
 	ctxWithResp := runtime.WithCaptureResponse(ctx, &rawResponse)
 
-	createSubnet := true
+	updateSubnet := false
```
Fei-Guo, line 131: A better version would be
var updateSubnet, createNewSubnet bool

They will be initialized to 0. which is false.

pkg/network/aci_network.go
smritidahal653: updateSubnet needs to start out true because shouldCreateSubnet only toggles to false but I like [this suggestion](https://github.com/virtual-kubelet/azure-aci/pull/628#discussion_r1699259633) for readability that's why I set the updateSubnet to true here.

smritidahal653: updated

Fei-Guo (approved):

PR #634, opened 2024-08-08
fix: bugfixes and release condition update

smritidahal653:
- Subnet [PUT](https://learn.microsoft.com/en-us/rest/api/virtualnetwork/subnets/create-or-update?view=rest-virtualnetwork-2024-01-01&tabs=HTTP) findings for addressPrefix vs addressPrefixes usage:
  - If neither is included in the payload, the request will return an error with code "AddressPrefixStringCannotBeNullOrEmpty"
  - If we send in payload with addressPrefix, resource will have properties.addressPrefix populated
  - If we send in payload with addressPrefixes, resource will have properties.addressPrefixes populated
  - If we send in payload with both addressPrefix and addressPrefixes, resource will have properties.addressPrefix populated (even if addressPrefix and addressPrefixes have two different values)
- Changes to aci_network.go: updated CreateOrUpdateACISubnet function so it takes the currentSubnet as a parameter and checks whether its a subnet create vs update flow based on it (moved this logic out of setupNetwork function to keep it simple). We only set the AddressPrefix on create and ACI service delegation is added to the payload in both create and update scenarios.
- fixing init validation container test to use the vk pod as aci-connector-linux pod is only created on --enable-addon flow 
- fixing workflow triggers to match current release branch naming convention

### DIFF
--- a/pkg/network/aci_network.go
+++ b/pkg/network/aci_network.go
@@ -134,12 +134,12 @@ func (pn *ProviderNetwork) setupNetwork(ctx context.Context, azConfig *auth.Conf
 	}
 
 	// check if the current subnet is valid or if we need to create a new subnet
-	updateSubnet, err := pn.shouldCreateSubnet(currentSubnet)
+	createOrUpdateSubnet, err := pn.shouldCreateOrUpdateSubnet(currentSubnet)
 	if err != nil {
 		return err
 	}
 
-	if updateSubnet {
+	if createOrUpdateSubnet {
 		// decide whether to create a new subnet or update the existing one based currentSubnet
 		err2 := pn.CreateOrUpdateACISubnet(ctx, subnetsClient, currentSubnet)
 		if err2 != nil {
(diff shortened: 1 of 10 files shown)


pkg/network/aci_network.go
```diff
@@ -280,7 +281,6 @@ func (pn *ProviderNetwork) CreateOrUpdateACISubnet(ctx context.Context, subnetsC
 
 	// only set the address prefix and prefixes if we are creating a new subnet
 	if isCreate {
-		subnet.Properties.AddressPrefix = &pn.SubnetCIDR
```
helayoty, line 283: why do we need to remove it in the isCreate case?

smritidahal653: Since Properties.AddressPrefix is required in the PUT payload, I am moving it out of just the create block and moving it up for both create and update scenarios

Fei-Guo (approved):

PR #680, opened 2024-11-14
fix: exclude confidential container test

smritidahal653:
Confidential pods are stuck in pending due to a regression with Plan9 mounts in the containerd package, causing the confidential container test to fail. Exluding this test from our e2e tests till the issue is resolved.


Makefile
```diff
@@ -105,15 +105,15 @@ e2e-test:
  	IMG_URL=$(REGISTRY) IMG_REPO=$(IMG_NAME) IMG_TAG=$(IMG_TAG) \
  	INIT_IMG_REPO=$(INIT_IMG_NAME) INIT_IMG_TAG=$(INIT_IMG_TAG) \
  	LOCATION=$(LOCATION) RESOURCE_GROUP=$(E2E_CLUSTER_NAME) \
- 	$(AKS_E2E_SCRIPT) go test -timeout 90m -v ./e2e
+ 	$(AKS_E2E_SCRIPT) go test -timeout 90m -v $(shell go list ./e2e/... | grep -v /e2e/confidential_container_test.go)
```
Fei-Guo, line 108: Please add a comment about why the test is excluded.

Fei-Guo (approved):

